PR #16, opened 2014-11-14
[NO_MERGE] [JENKINS-25612] - PropertyFileCredentials

oleg-nenashev:
This is a draft implementation for the feasibility elaboration. An input from plugin owners would be useful.

TODOs:
- Fields validation (in particular, security notifications)
- Help messages
- Unit tests


stephenc:
> Yes, there may be problems with distributed Jenkins installations if somebody uses a local file

That's what [CredentialsProvider.snapshot(Credentials)](https://github.com/jenkinsci/credentials-plugin/blob/master/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java#L555) is for. You use that method before sending the credential to a slave

PR #21, opened 2015-01-13
[FIXED JENKINS-26099] Pick an ID for credentials

jglick:
[JENKINS-26099](https://issues.jenkins-ci.org/browse/JENKINS-26099)

This PR only covers credentials types defined in this plugin. Can create follow-on PRs for `BasicSSHUserPrivateKey`, `StringCredentialsImpl`, and `FileCredentialsImpl`.

- [x] deal with `fixInstance` hack
- [x] verify that IDs are using permitted characters (which?)
- [x] verify that IDs do not clash with those of existing credentials
- [x] only search existing credentials this user can see
- [x] differentiate existing credentials so that those in other stores would be `WARNING` only

@reviewbybees


jglick:
And @stephenc can the workarounds for [JENKINS-19124](https://issues.jenkins-ci.org/browse/JENKINS-19124), supposedly fixed in 1.528, be deleted at this point (if we update the baseline to 1.532 or 1.532.x)?

jglick:
Whatever the JENKINS-19124 workaround is supposed to be doing, it is not making `FileOnMasterKeyStoreSource.DescriptorImpl.doCheckKeyStoreFile` work (in `master`): `password` is set correctly at first, and form validation runs, but then the check method is called several more times with `password == null`. I will see if I can fix this.

src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java
```diff
@@ -109,4 +117,35 @@ public final boolean equals(Object o) {
...
+
+        public final FormValidation doCheckId(@QueryParameter String value) {
+            if (value.isEmpty()) {
+                return FormValidation.ok();
+            }
+            if (!value.matches("[a-zA-Z0-9_.-]+")) { // anything else considered kosher?
+                return FormValidation.error("Unacceptable characters");
+            }
+            // TODO perhaps limit to the store associated with a User or ItemGroup (Jenkins/Folder) in @AncestorInPath
+            if (CredentialsMatchers.firstOrNull(CredentialsProvider.lookupCredentials(IdCredentials.class, (ItemGroup<?>) null, null, Collections.<DomainRequirement>emptyList()), CredentialsMatchers.withId(value)) != null) {
```
stephenc: Also may want to check current user's credentials and probably should be limited to the store that this credential is being created within (as you could use a fixed id to let different users use their own credentials be picked up by the job)... so only a duplicate in the current store should be an ERROR. a duplicate globally should be a WARNING

stephenc:
The doCheckId code needs some work, otherwise it's looks good to go

src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java
jglick: > may want to check current user's credentials

Probably right.

> should be limited to the store that this credential is being created within […]
> a duplicate in the current store should be an ERROR. a duplicate globally should be a WARNING

Agreed, this would be ideal. I will probably need help with the API here. I know how to get a `User` or `ItemGroup` from the request, but am not sure how to select the corresponding `CredentialsStore` (especially when inheritance of credentials from parent folders/root is considered).

stephenc: You just iterate the CredentialsProvider.lookupStores(ModelObject). The first one is the one where duplicates are an ERROR, the remaining ones just give you a WARNING. Then finally if the ModelObject is not a User you would need to check the User

stephenc: I'd need to check if you need to add a special case for old-style the ManagementLink UI but that should pick up Jenkins as the ModelObject and thus work also

jglick: > You just iterate the CredentialsProvider.lookupStores(ModelObject)

I will try it and see.

pom.xml
```diff
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.520</version>
+    <version>1.526</version>
```
jglick: To match baseline of `cloudbees-folder`.

pom.xml
```diff
@@ -120,6 +120,12 @@
       <version>1.9.5</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>cloudbees-folder</artifactId>
+      <version>4.2.3</version>
+      <scope>test</scope>
+    </dependency>
```
jglick: To include `Folder`s in test. Unfortunately this seems to have exposed a subtle bug in `maven-hpi-plugin`: we would like `hpi:run` to create `work/plugins/cloudbees-folder.jpi`, which it does, and `target/credentials.hpi!/META-INF/MANIFEST.MF` and `target/test-classes/the.hpl` both correctly omit mention of this dependency (because the scope is `test`), yet `work/plugins/credentials.hpl` includes

    Plugin-Dependencies: ui-samples-plugin:1.526,cloudbees-folder:4.2.3

which causes `hpi:run` to produce the error

```
… hudson.PluginManager$1$3$2$1 reactOnCycle
SEVERE: found cycle in plugin dependencies: (root=Plugin:credentials, deactivating all involved) Plugin:credentials -> Plugin:cloudbees-folder -> Plugin:credentials
```

and refuse to load folder-related extensions.

This seems to have been caused by https://github.com/jenkinsci/maven-hpi-plugin/commit/20e49d0 in combination with https://github.com/jenkinsci/maven-hpi-plugin/commit/7ea2da1. It is unfortunate that the ITs do not test `hpi:run` at all.

stephenc:
If you update the warning message to tell the user where there is a conflict then I think this is good to go

jglick:
@stephenc again please comment on the PR diff page, not on individual comments.

stephenc:
+1 and release away

PR #23, opened 2015-01-22
Adjuncts do not include CSS correctly for removable dom fragments

stephenc:
- I think it only works by accident with script tags
- When the adjunct is initially included the script and css elements are injected at the point where the st:adjunct
  tag is first referenced.
- Thus these tags form part of the DOM that can be removed (e.g. when part of a repeatable / hetero list)
- Because the script is executed, removing the DOM node cannot un-execute the script, so script tags are safe
- Removing the css's link tag, however, removes the styling... and the adjunct manager doesn't know.
- What this hack does is inject the CSS at the end of the DOM's head tag (yes triggering a reflow) but only
  if the CSS link is not present already. The JavaScript tag (because we are bypassing st:adjunct) is conditionally
  appended to the DOM's body so that it can be fetched while other scripts continue executing.

To see the bug this fixes:

- Create a job with the GIT SCM and a couple of SCM locations and some credential parameters
- Save the job
- Load the job
- Remove the first credential parameter, now all the credential tags loose the styling (width:auto on the drop-down) and the Add button overflows to the next line

### DIFF
--- a/src/main/resources/lib/credentials/select.jelly
+++ b/src/main/resources/lib/credentials/select.jelly
@@ -43,7 +43,6 @@
       If expressions are allowed then the user can choose between a select drop down and entering an expression.
     </st:attribute>
   </st:documentation>
-  <st:adjunct includes="lib.credentials.select.select"/>
   <f:prepareDatabinding/>
   <j:set var="value" value="${attrs.value ?: instance[attrs.field] ?: attrs.default}" />
   <j:set var="paramValue" value="${attrs.expressionAllowed and value.startsWith('${') and value.endsWith('}')}"/>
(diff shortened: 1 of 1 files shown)


stephenc:
@kohsuke here's a strange one for you!

stephenc:
@reviewbybees

jglick:
Ideally this could be proven with a functional test, but I am not sure if HtmlUnit is powerful enough to deal with the dynamic DOM changes.

This should be fixed upstream in Stapler.

kohsuke:
Created [JENKINS-26578](https://issues.jenkins-ci.org/browse/JENKINS-26578) to track the fix in stapler. A comment should be added that this fix is a temporary work around until that gets resolved.

I'm OK with pushing this fix in, until core picks up JENKINS-26578 whenever that happens.

rsandell:
+1

PR #28, opened 2015-07-02
Update baseline Jenkins version and fix corresponding findbugs issues

stephenc:


stephenc:
@reviewbybees

oleg-nenashev:
:bee:

src/main/java/com/cloudbees/plugins/credentials/BaseCredentials.java
```diff
@@ -76,6 +76,11 @@ public CredentialsScope getScope() {
     @NonNull
     @SuppressWarnings("unchecked")
     public CredentialsDescriptor getDescriptor() {
-        return (CredentialsDescriptor) Hudson.getInstance().getDescriptorOrDie(getClass());
+        // TODO switch to Jenkins.getActiveInstance() once 1.590+ is the baseline
+        Jenkins jenkins = Jenkins.getInstance();
+        if (jenkins == null) {
+            throw new IllegalStateException("Jenkins has not been started, or was already shut down");
+        }
```
amuniz: Wasn't there ```Jenkins.getActiveInstance()``` in 1.565, right?

src/main/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinition.java
```diff
@@ -135,7 +135,12 @@ private boolean match(Set<Class<? extends StandardCredentials>> allowed, Standar
 
         public StandardListBoxModel doFillDefaultValueItems(@AncestorInPath Item context,
                                                             @QueryParameter(required = true) String credentialType) {
-            final ACL acl = context == null ? Jenkins.getInstance().getACL() : context.getACL();
+            // TODO switch to Jenkins.getActiveInstance() once 1.590+ is the baseline
```
amuniz: Ok. This is answering the previous comment...

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -836,4 +807,12 @@ public void remove() {
         return null;
     }
 
+    public static ExtensionList<CredentialsProvider> all() {
+        // TODO switch to ExtensionList.lookup once Jenkins 1.572+ is the baseline
+        final Jenkins jenkins = Jenkins.getInstance();
+        return jenkins == null
+                ? ExtensionList.create((Jenkins)null, CredentialsProvider.class)
```
amuniz: Are you casting ```null``` to ```Jenkins```?

amuniz:
:bee: (the null casting is a somewhat strange to me, but not an issue, of course)

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
stephenc: Because there is a second method that has the `Hudson` as the first parameter so when using a `null` you need to tell the compiler the type to use... we could use the `jenkins` variable but that is a known `null` so code analysis complains about that too

stephenc:
@reviewbybees done

stephenc:
Wearing my plugin maintainer's hat, I intend merging this tomorrow unless I hear otherwise

PR #32, opened 2015-09-22
WIP credentials: add CredentialSet for strict domain->credentials map

jacob-keller:
This is based off of the submodules pull request in git-plugin, and @stephenc's ideas about how to provide credentials to the user.

I am not sure if this fully does what we expect here, but I think it's a starting point to discuss. I really need help with the Jelly side of things, and exactly how to present these to the user.

I've removed all the code except for a single constructor. This now implements from BaseStandardCredentials, ie: it is itself a credentials.

I think we need a "resolveCredentials" method which can be used to resolve some domain requirements into the credential from the set that matches. We won't expose the map contract, but the default constructor will be used to create the set from a map.

This is *not* ready to merge at this time.

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/impl/CredentialsSet.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/impl/CredentialsSet.java
@@ -1,0 +1,62 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 Jacob Keller
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.cloudbees.plugins.credentials.impl;
+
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/plugins/credentials/domains/CredentialSet.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.TreeMap;
+import java.util.Map;
+import java.util.List;
+
+/**
+ * Represents a mapping between {@link Domain}s and {@link Credentials},
+ * similar to {@link DomainCredentials}, but providing exact credentials for
+ * each domain.
+ */
+public class CredentialSet extends TreeMap<Domain, Credentials> {
```
stephenc: So I don't see it as a map between Domains and Credentials... rather a `Map<Set<DomainSpecification>,Credential>`. I also find inheriting from `TreeMap` as risky... for one the on-disk format should probably be `Map<Set<DomainSpecification>,String>` i.e. persisting the credential id rather than the credential itself...

And once you head down that road you actually probably want to just hide the map contract entirely and just give people a helper method... IOW you want to have them do

```
CredentialSet set = ...;
...
FooCredentials fooCreds = set.resolveCredentials(FooCredentials.class, context, auth, domainRequirements);
```

I suspect we want to flesh out some use cases as integrators would want first before trying to code this up

jacob-keller: Ok. I like the idea of dropping the map contract.

Another alternative is to actually extend the BaseCredentials so that a CredentialsSet could actually be treated like a complete set.

What's the difference between a set of DomainSpecifications and a Domain?

I don't really know if we want to extend credentials because I don't think it really honors the domain in the same way.. ie.. you can't really put credentialsSet inside a single domain... but then how does one perform lookups for which set you're going to use?

Generally, I think the use case of a CredentialsSet is something like:

"Given this domain, use this credentials" vs the Domain which is "here is a list of credentials you can choose from if you match this domain".

This is a stronger contract and easier to use. In addition, a CredentialSet enables a collection of domains and their given credentials.

I think the primary connection to them from the use side of credentials should be the "resolveCredentials" as above.

Makes sense to use the string instead of the credentials object itself.

The main reason I implemented it as a contract of a map was to make it easier for the supplying side of credentials to add and modify the mappings.. But maybe that's not really important... We can simply provide a constructor instead that takes a Map.

jacob-keller:
I definitely need some more help/input on this one...

stephenc:
I have a design in my head. There are issues with this line of attack that you are following. I wish I had more time to explain the issues I see. Hopefully I will get some time in the next week or two to flesh out the design I have in mind

jacob-keller:
Ok. I will be patient. I read what you have in the other plugin as a base
but I am missing something I think.

jacob-keller:
Just a friendly reminder about this.

AnneTheAgile:
+1 :)
As I see it; 
1.The Jenkins Git plugin ticket is not solved anywhere yet, the best I can see. The v2.5 experimental patch was not accepted, is no longer on the beta, and the last release is 2.4.4. Its PR was closed, not implemented in favor of an edit to the credentials plugin. 
https://issues.jenkins-ci.org/browse/JENKINS-20941
https://github.com/jenkinsci/git-plugin/pull/342

2.The Jenkins Credentials plugin PR has similarly not been accepted (but it needs work so is not standalone). 
https://github.com/jenkinsci/credentials-plugin/pull/32

stephenc:
I am closing this PR as it is not the direction I want to take solving this problem. Thank you very much for your contribution.

PR #33, opened 2015-09-28
Implemented prototype Rest API for modifying domains and credentials

roccivic:
Hi,

I'm interested in having REST API support for managing domains and credentials, which seems to have been discussed previously:
https://issues.jenkins-ci.org/browse/JENKINS-28407
https://groups.google.com/forum/#!topic/jenkinsci-dev/BCZr7Bn9sKY

Unfortunately I'm a newbie when it comes to Jenkins development, so I'm pretty sure what I'm proposing in the pull request is wrong.
I hope someone could point me in the right direction for implementing this feature properly.

What I've attempted to do is:

* Add some notes on how to use the API using jelly views
* Respond to HTTP requests with a status code which should be meaningful alone, instead of a 302 Redirect
* Respond with the ID of the created credential
* Take the incoming data directly from the payload as a JSON object, instead of an element from an x-www-form-urlencoded payload

The open questions for me currently are:
* The urls that I chose are probably strange. E.g.: ```apiCreateCredentials```. How could I make these better?
* There is some code duplication between the API methods and their regular counterparts. How could I reduce this duplication?


src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/CredentialsWrapper/_api.jelly
```diff
@@ -0,0 +1,65 @@
...
+    </p>
+    <h2>Update credentials</h2>
+    <p>
+        POST to <a href="../apiUpdateCredentials"><code>apiUpdateCredentials</code></a> to update credentials.
+        Use the following data as a payload:
+    </p>
+    <p>
+        <code>{<br/>
+            "description": "This is the description for my credentials",<br/>
+            "stapler-class": "com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl",<br/>
```
jglick: `$class` should be used in recent versions of Jenkins.

src/main/java/com/cloudbees/plugins/credentials/CredentialsStoreAction.java
```diff
@@ -543,6 +599,18 @@ public HttpResponse doUpdateSubmit(StaplerRequest req) throws ServletException,
             return HttpResponses.redirectToDot();
         }
 
+        @RequirePOST
+        @WebMethod(name = "apiUpdateCredentials")
```
jglick: `name` is redundant—inferred from the method name.

src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/CredentialsWrapper/_api.jelly
```diff
@@ -0,0 +1,65 @@
...
+ -->
+
+<!--
+  Index page
+-->
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core">
+    <h2>Delete credentials</h2>
+    <p>
+        POST to <a href="../apiDeleteCredentials"><code>apiDeleteCredentials</code></a> to delete these credentials.
```
jglick: Why the `api` prefix? Just `deleteCredentials` would be better. Or even `delete`.

roccivic:
I was able to call the existing endpoints without too much hastle, so closing this for now as the pull request appears to be redundant.

dcoraboeuf:
@roccivic, as a matter of fact, how did you finally proceed?

roccivic:
@dcoraboeuf Well, I had a few issues to work around, but nothing major:
* Existing endpoints only reply with a 302 redirect
  * Solved with a fire-and-forget approach, with a second request to see if the object was created/deleted.
* The ID of the created object is not returned
  * Solved by providing a GUID as the new ID from my side.
* I wanted the request payload to go up as ```application/json```
  * Just used the standard Jenkins way of adding the JSON payload to a url-encoded form.

All that I really needed was to know the exact payload to send in the request for the various actions. I just used the network tab of the developer tools in chrome to watch the request as I performed the action in Jenkins.
Authenticating is pretty much the same as for other endpoints: Basic authentication using the user's API token.

I can't share the code as it's from a closed source application, but let me know if I can help any further.

nicorikken:
It seems that this python script does the job: https://github.com/wunzeco/ansible-jenkins-extra/blob/master/files/create-credentials-config.py as called in the corresponding Ansible script https://github.com/wunzeco/ansible-jenkins-extra/blob/master/tasks/job-creds.yml Going to check it out right now.

nicorikken:
I can confirm that the ansible-jenkins-extra role sets credentials nicely, even though I've only tested the Jenkins job credentials. The repo could use some love in the little details, so please give it a look and give it a go.

PR #38, opened 2015-12-24
Add icon name to descriptor

halkeye:
Merged in icon-tag-shim-test and then extended it to include icon class name in the descriptor.

I read over the contributing guidelines. I can't see anything not setup right.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,6 +106,11 @@
       <version>1.3.9</version>
     </dependency>
     <!-- plugin dependencies -->
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.icon-shim</groupId>
+      <artifactId>icon-shim</artifactId>
+      <version>1.0.3</version>
+    </dependency>
     <!-- jenkins dependencies -->
     <!-- test dependencies -->
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -109,7 +109,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins.icon-shim</groupId>
       <artifactId>icon-shim</artifactId>
-      <version>1.0.3</version>
+      <version>2.0.2</version>
     </dependency>
     <!-- jenkins dependencies -->
     <!-- test dependencies -->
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -106,6 +106,11 @@
       <version>1.3.9</version>
     </dependency>
     <!-- plugin dependencies -->
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.icon-shim</groupId>
+      <artifactId>icon-shim</artifactId>
+      <version>1.0.3</version>
```
tfennelly: Seems like `2.0.1` is the latest release. Might be good to try with that.

tfennelly:
@halkeye LGTM.

Thanks for fixing this up. Not sure why we didn't merge the original commit. I see no reason not to do it.

halkeye:
@tfennelly Updating to 2.0.1 seems to break things. 
I changed the pom file, ran hpi:run and have no images.

I'm thinking it has something todo with https://github.com/jenkinsci/icon-shim-plugin/compare/icon-shim-pom-1.0.3...icon-shim-pom-2.0.1#diff-c0aec591762b9fcb07f322e6bb962df5L95

Is there something I need to do to get the icons into css now?

halkeye:
@tfennelly infact the readme on still says to use 1.0.3 (https://github.com/jenkinsci/icon-shim-plugin)

halkeye:
Yea, looks like 2.0 is a whole different design, but I could update to latest 1.0, but the readme for the repo/module says 1.0.3 so was kinda sticking with it.

Pinging @stephenc as he looks like the current maintainer for a review

halkeye:
Okay, updated for 2.0.2 as @tfennelly updated it to make the shim work properly again.

oleg-nenashev:
LGTM :+1:

src/main/java/com/cloudbees/plugins/credentials/PluginImpl.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2004-2009, Sun Microsystems, Inc., Kohsuke Kawaguchi
```
KostyaSha: copy-paste?

halkeye: you'll have to ask @tfennelly thats his patch

PR #39, opened 2016-01-19
CredentialsParameterDefinition.getDefaultParameterValue().isDefaultValue() should return true

mikedougherty:
This causes problems with credentials-binding-plugin failing to find a credentials value when a job is started by System (e.g. timer, SCM push, etc).

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/credentials-plugin/39)


daniel-beck:
Like an idiot I submitted #40 without looking at existing PRs. My apologies.

Since I filed an issue and have a test, I'd prefer mine over this. WDYT?

mikedougherty:
Having the test is much better, I just didn't have a good java dev environment. Let's go with yours, @daniel-beck

PR #40, opened 2016-01-27
[FIX JENKINS-32642] Consider default value to be the default

daniel-beck:
Not sure whether this is in fact the intended use of `isDefaultValue()`, but the lack of Javadoc meant I had to guess. This PR may well be misguided.

https://issues.jenkins-ci.org/browse/JENKINS-32642

@reviewbybees

### DIFF
--- a/src/test/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinitionTest.java
+++ b/src/test/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinitionTest.java
@@ -1,0 +1,83 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2016 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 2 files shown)


src/test/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinitionTest.java
```diff
@@ -0,0 +1,83 @@
...
+        List<ParameterDefinition> params = new ArrayList<ParameterDefinition>();
+        params.add(new CredentialsParameterDefinition("name", "description", "id", "type", true));
+
+        ParametersDefinitionProperty prop = new ParametersDefinitionProperty(params);
+        p.addProperty(prop);
+        p.getBuildersList().add(new ParamCheckBuilder());
+        r.buildAndAssertSuccess(p);
+    }
+
+    public static class ParamCheckBuilder extends Builder {
```
oleg-nenashev: NIT: Could have a more specific name

src/test/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinitionTest.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.List;
+
+import static org.junit.Assert.*;
+
+public class CredentialsParameterDefinitionTest {
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    @Test public void defaultValue() throws Exception {
```
oleg-nenashev: Would be nice to add the issue reference

oleg-nenashev:
The change is reasonable IMHO, so :+1: 
I would restrict the 3-parameter constructor of <code> CredentialsParameterValue</code> to avoid such issues in the future

PR #42, opened 2016-03-09
[JENKINS-31863] Sort Credential select list by credenial name

AndraeRay:
Implemented comparator to sort credentials by credential name in select lists. For simplicity and safety I only modified the returned ArrayList and not any of the add credential functions or the main display page.

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
@@ -840,6 +840,10 @@ public abstract class CredentialsProvider implements ExtensionPoint {
 
 }
 
+/**
+ * Comparator to alphabetically sort credentials drop down list in ascending order by Credential Name.
+ *
+ */
 class CredentialsNameComparator implements Comparator<Credentials>{
 	 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -834,3 +839,15 @@ public void remove() {
     }
 
 }
+
+/**
+ * Comparator to alphabetically sort credentials drop down list in ascending order by Credential Name.
+ *
+ */
+class CredentialsNameComparator implements Comparator<Credentials>, java.io.Serializable{
+	private static final long serialVersionUID = 1;
```
halkeye: I thought serialVersionUID was no longer needed in "modern" java

halkeye: Also there seems to be some indenting differences here

AndraeRay: My eclipse gave me a note about it so I wasn't sure if findbugs would also mention it.

whoops i'll fix whitespace.

PR #51, opened 2016-05-25
[FIXED JENKINS-35130] Delete and Move credentials do not work with providers that do not have modifiable domains

stephenc:
@reviewbybees see [JENKINS-35130](https://issues.jenkins-ci.org/browse/JENKINS-35130)


jglick:
:bee:

PR #52, opened 2016-05-27
[FIXED JENKINS-35168] Resolve the context `ModelObject` when `it` isn't a `ModelObeject`

stephenc:
@reviewbybees

### DIFF
--- a/src/main/resources/lib/credentials/select.jelly
+++ b/src/main/resources/lib/credentials/select.jelly
@@ -45,6 +45,9 @@
     <st:attribute name="includeUser">
       If the user's scoped stores should be included
     </st:attribute>
+    <st:attribute name="context">
+      The context within which to look up the stores. (defaults to <code>it</code> (jelly) / <code>my</code> (groovy))
+    </st:attribute>
   </st:documentation>
   <f:prepareDatabinding/>
   <j:set var="value" value="${attrs.value ?: instance[attrs.field] ?: attrs.default}" />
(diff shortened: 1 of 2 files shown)


src/main/resources/lib/credentials/select.jelly
```diff
@@ -71,8 +75,9 @@
                 default="${paramDefault?'':attrs.default}" value="${paramValue?(paramDefault?'':attrs.default):value}"/>
       <!-- TODO  add support for checking permissions against stores in request path -->
       <st:nbsp/>
+      <j:set var="storeItems" value="${selectHelper.getStoreItems(context, includeUser)}"/>
       <j:choose>
-        <j:when test="${h.hasPermission(selectHelper.CREATE)}">
+        <j:when test="${h.hasPermission(selectHelper.CREATE) and storeItems != null and !storeItems.isEmpty()}">
```
tfennelly: I didn't know `and` would work in those expressions :)

tfennelly:
🐝

src/main/resources/lib/credentials/select.jelly
oleg-nenashev: Seems to be another unrelated fix, but LGTM

oleg-nenashev:
🐝

src/main/resources/lib/credentials/select.jelly
stephenc: @tfennelly yes it is a special alias because typing `&amp;&amp;` would be a pain

stephenc: @oleg-nenashev not quite, if the resolution fails we will have an empty list so the button should be disabled

PR #53, opened 2016-05-27
[JENKINS-35150] We can be more forgiving of users with misconfigured HTTPS front-ends

stephenc:
See [JENKINS-35150](https://issues.jenkins-ci.org/browse/JENKINS-35150)

@reviewbybees


tfennelly:
🐝

oleg-nenashev:
Even if we forgive them, other routines won't do so...
But 🐝

stephenc:
@oleg-nenashev this is true, but we have the ability to forgive them, so we might as well

PR #54, opened 2016-05-31
[FIXED JENKINS-24631] Expose the credentials details by the XML/JSON API

stephenc:
@reviewbybees


recena:
👍  and 🐝

jglick:
:bee: but it really deserves a test. Should be easy to write with `WebClient`.

PR #55, opened 2016-06-02
[JENKINS-35306] Credentials Providers need to be able to list credentials without retrieving the backing secret

stephenc:
See [JENKINS-35306](https://issues.jenkins-ci.org/browse/JENKINS-35306)

@reviewbybees


stephenc:
(findbugs is not failing locally....)

PR #57, opened 2016-06-13
[JENKINS-28407] Create Credentials through CLI

stephenc:
[JENKINS-28407](https://issues.jenkins-ci.org/browse/JENKINS-28407)

- [x] CLI commands for CRUD of credentials
- [x] CLI commands for CRUD of credential domains
- [x] CLI commands to list credential stores
- [x] REST support for CRU of credentials
- [x] REST support for CRU of credential domains
- [x] Decide if we want REST support for D of credentials and credential domains
- [x] Happy path tests: C for credentials and credential domains via CLI
- [x] Happy path tests: R for credentials and credential domains via CLI
- [x] Happy path tests: U for credentials and credential domains via CLI
- [x] Happy path tests: D for credentials and credential domains via CLI
- [x] Happy path tests: C for credentials and credential domains via REST
- [x] Happy path tests: R for credentials and credential domains via REST
- [x] Happy path tests: U for credentials and credential domains via REST
- [x] Happy path tests: D for credentials and credential domains via REST
- [x] Happy path tests: listing credential stores via CLI
- [x] Some non-happy path tests

PR #58, opened 2016-06-14
[JENKINS-20139] Add support to track where a credential is used

stephenc:
[JENKINS-20139](https://issues.jenkins-ci.org/browse/JENKINS-20139)

- Needs to wait for https://github.com/jenkinsci/credentials-plugin/pull/57 to be merged before we can call this done (see TODO task in code)

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsParameterValue.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsParameterValue.java
@@ -9,12 +9,14 @@ import hudson.EnvVars;
 import hudson.Util;
 import hudson.model.AbstractBuild;
 import hudson.model.Executor;
+import hudson.model.Fingerprint;
 import hudson.model.ParameterValue;
 import hudson.model.Run;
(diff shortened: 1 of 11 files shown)


stephenc:
screen shot 2016-06-14 at 15 32 19

halkeye:
Peeking at the code, is this something that plugin developers will have to add? 
`CredentialsProvider.trackUsage(build, lookupCredentials(StandardCredentials.class, build));`

stephenc:
Not quite, when a plugin developer retrieves a credential for actual **use** (distinct from pre-validation, probing, etc) they will need to call one of the track methods (or do their own facets if doing something more fancy)

I have no way of distinguishing generic retrieval from usage, hence that necessity. Hopefully I can update the wiki with recommended integration patterns before I get pulled away again

stephenc:
`CredentialsProvider.trackUsage(build, credential)` is the intended usage. Credential parameters are different as they do not normally actually retrieve the credential, rather just the credential ID which is why I need to look it up for tracking.

PR #62, opened 2016-07-04
[FIXED JENKINS-36315] Make it easier to infer the context in form binding

stephenc:
[JENKINS-36315](https://issues.jenkins-ci.org/browse/JENKINS-36315)

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/ContextInPath.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/ContextInPath.java
@@ -1,0 +1,46 @@
+package com.cloudbees.plugins.credentials;
+
+import hudson.model.ModelObject;
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import javax.servlet.ServletException;
+import org.apache.commons.lang.StringUtils;
+import org.kohsuke.stapler.AnnotationHandler;
+import org.kohsuke.stapler.InjectedParameter;
+import org.kohsuke.stapler.StaplerRequest;
+
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Indicates that this parameter is injected by evaluating
+ * {@link StaplerRequest#getAncestors()} and searching for a credentials context with the parameter type.
+ * You can enhance the lookup by ensuring that there are query parameters of {@code $provider} and {@code $token}
+ * that correspond to the context's {@link CredentialsSelectHelper.ContextResolver} FQCN and
+ * {@link CredentialsSelectHelper.ContextResolver#getToken(ModelObject)} respectively.
+ *
+ * @see CredentialsDescriptor#getCheckMethod(String)
+ * @since 2.1.5
+ */
+@Retention(RUNTIME)
+@Target(PARAMETER)
+@Documented
+@InjectedParameter(ContextInPath.HandlerImpl.class)
+public @interface ContextInPath {
+    class HandlerImpl extends AnnotationHandler<ContextInPath> {
+        public Object parse(StaplerRequest request, ContextInPath contextInPath, Class type, String parameterName)
+                throws
+                ServletException {
+            String $provider = request.getParameter("$provider");
+            String $token = request.getParameter("$token");
+            if (StringUtils.isNotBlank($provider) && StringUtils.isNotBlank($token)) {
+                ModelObject context = CredentialsDescriptor.lookupContext($provider, $token);
+                if (type.isInstance(context)) {
+                    return type.cast(context);
+                }
+            }
+            return CredentialsDescriptor.findContextInPath(request, type);
+        }
+    }
+}
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
@@ -92,12 +92,12 @@ public abstract class CredentialsDescriptor extends Descriptor<Credentials> impl
     @RestrictedSince("2.1.5")
     public ListBoxModel doFillScopeItems(@ContextInPath ModelObject context) {
         ListBoxModel m = new ListBoxModel();
-                Set<CredentialsScope> scopes = CredentialsProvider.lookupScopes(context);
-                if (scopes != null) {
-                    for (CredentialsScope scope : scopes) {
-                        m.add(scope.getDisplayName(), scope.toString());
-                    }
-                }
+        Set<CredentialsScope> scopes = CredentialsProvider.lookupScopes(context);
+        if (scopes != null) {
+            for (CredentialsScope scope : scopes) {
+                m.add(scope.getDisplayName(), scope.toString());
+            }
+        }
         return m;
     }
 
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
@@ -23,19 +23,44 @@
  */
 package com.cloudbees.plugins.credentials;
 
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.ExtensionList;
+import hudson.Functions;
+import hudson.RestrictedSince;
 import hudson.model.Descriptor;
-import hudson.model.DescriptorVisibilityFilter;
+import hudson.model.Item;
 import hudson.model.ModelObject;
+import hudson.model.User;
+import hudson.util.FormValidation;
 import hudson.util.ListBoxModel;
+import java.io.UnsupportedEncodingException;
+import java.net.URI;
+import java.net.URLEncoder;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import jenkins.model.Jenkins;
+import org.apache.commons.jelly.JellyContext;
+import org.apache.commons.lang.StringUtils;
 import org.jenkins.ui.icon.IconSpec;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.Ancestor;
+import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.Stapler;
+import org.kohsuke.stapler.StaplerRequest;
 
 /**
  * Descriptor for credentials.
  */
 public abstract class CredentialsDescriptor extends Descriptor<Credentials> implements IconSpec {
+
+    private transient final Map<String, FormValidation.CheckMethod>
+            enhancedCheckMethods = new ConcurrentHashMap<String, FormValidation.CheckMethod>();
 
     /**
      * Constructor.
@@ -63,22 +88,16 @@ public abstract class CredentialsDescriptor extends Descriptor<Credentials> impl
      * @return the scopes for the nearest request object that acts as a container for credentials.
      */
     @SuppressWarnings("unused") // used by stapler
-    public ListBoxModel doFillScopeItems() {
+    @Restricted(NoExternalUse.class)
+    @RestrictedSince("2.1.5")
+    public ListBoxModel doFillScopeItems(@ContextInPath ModelObject context) {
         ListBoxModel m = new ListBoxModel();
-        Ancestor ancestor = Stapler.getCurrentRequest().findAncestor(Object.class);
-        while (ancestor != null) {
-            if (ancestor.getObject() instanceof ModelObject) {
-                Set<CredentialsScope> scopes =
-                        CredentialsProvider.lookupScopes((ModelObject) ancestor.getObject());
+                Set<CredentialsScope> scopes = CredentialsProvider.lookupScopes(context);
                 if (scopes != null) {
                     for (CredentialsScope scope : scopes) {
                         m.add(scope.getDisplayName(), scope.toString());
                     }
-                    break;
                 }
-            }
-            ancestor = ancestor.getPrev();
-        }
         return m;
     }
 
@@ -93,12 +112,14 @@ public abstract class CredentialsDescriptor extends Descriptor<Credentials> impl
      * to be specified.
      */
     @SuppressWarnings("unused") // used by stapler
+    @Restricted(NoExternalUse.class)
+    @RestrictedSince("2.1.5")
     public boolean isScopeRelevant() {
         Ancestor ancestor = Stapler.getCurrentRequest().findAncestor(Object.class);
         while (ancestor != null) {
             if (ancestor.getObject() instanceof ModelObject) {
-                Set<CredentialsScope> scopes =
-                        CredentialsProvider.lookupScopes((ModelObject) ancestor.getObject());
+                ModelObject context = unwrapContext((ModelObject) ancestor.getObject());
+                Set<CredentialsScope> scopes = CredentialsProvider.lookupScopes(context);
                 if (scopes != null) {
                     return scopes.size() > 1;
                 }
@@ -117,8 +138,26 @@ public abstract class CredentialsDescriptor extends Descriptor<Credentials> impl
      * object.
      */
     @SuppressWarnings("unused") // used by stapler
+    @Restricted(NoExternalUse.class)
+    @RestrictedSince("2.1.5")
     public boolean isScopeRelevant(ModelObject object) {
         Set<CredentialsScope> scopes = CredentialsProvider.lookupScopes(object);
+        return scopes != null && scopes.size() > 1;
+    }
+
+    /**
+     * Similar to {@link #isScopeRelevant()} but operating on a specific {@link CredentialsStore} rather than trying to
+     * infer from the stapler request.
+     *
+     * @param store the object that is going to contain the credential.
+     * @return {@code true} if there is more than one {@link CredentialsScope} that can be used for the specified
+     * object.
+     * @since 2.1.5
+     */
+    @SuppressWarnings("unused") // used by stapler
+    @Restricted(NoExternalUse.class)
+    public boolean isScopeRelevant(@CheckForNull CredentialsStore store) {
+        Set<CredentialsScope> scopes = store == null ? null : store.getScopes();
         return scopes != null && scopes.size() > 1;
     }
 
@@ -130,6 +169,8 @@ public abstract class CredentialsDescriptor extends Descriptor<Credentials> impl
      * object.
      */
     @SuppressWarnings("unused") // used by stapler
+    @Restricted(NoExternalUse.class)
+    @RestrictedSince("2.1.5")
     public boolean isScopeRelevant(CredentialsStoreAction.DomainWrapper wrapper) {
         if (wrapper != null) {
             return isScopeRelevant(wrapper.getStore().getContext());
@@ -150,7 +191,31 @@ public abstract class CredentialsDescriptor extends Descriptor<Credentials> impl
      * object.
      */
     @SuppressWarnings("unused") // used by stapler
+    @Restricted(NoExternalUse.class)
+    @RestrictedSince("2.1.5")
     public boolean isScopeRelevant(CredentialsStoreAction.CredentialsWrapper wrapper) {
+        if (wrapper != null) {
+            return isScopeRelevant(wrapper.getStore().getContext());
+        }
+        CredentialsStoreAction action =
+                Stapler.getCurrentRequest().findAncestorObject(CredentialsStoreAction.class);
+        if (action != null) {
+            return isScopeRelevant(action.getStore().getContext());
+        }
+        return isScopeRelevant();
+    }
+
+    /**
+     * Similar to {@link #isScopeRelevant()} but used by {@link CredentialsSelectHelper}.
+     *
+     * @param wrapper the wrapper for the domain that is going to contain the credential.
+     * @return {@code true} if there is more than one {@link CredentialsScope} that can be used for the specified
+     * object.
+     * @since 2.1.5
+     */
+    @SuppressWarnings("unused") // used by stapler
+    @Restricted(NoExternalUse.class)
+    public boolean isScopeRelevant(CredentialsSelectHelper.WrappedCredentialsStore wrapper) {
         if (wrapper != null) {
             return isScopeRelevant(wrapper.getStore().getContext());
         }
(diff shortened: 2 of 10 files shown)


src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,339 @@ public String getIconClassName() {
...
+                    }
+                    if (pathSegments.size() >= 2) {
+                        String firstSegment = pathSegments.get(0);
+                        if ("user".equals(firstSegment)) {
+                            User user = User.get(pathSegments.get(1));
+                            if (type.isInstance(user) && CredentialsProvider.hasStores(user)) {
+                                // we have a winner
+                                return type.cast(user);
+                            }
+                        } else if ("job".equals(firstSegment) || "item".equals(firstSegment) || "view"
```
stephenc: @jglick you may be interested in this section as a better implementation of https://github.com/jenkinsci/workflow-step-api-plugin/blob/aabbdca1c3c1f1aacf346b01d95d3be467547195/src/main/java/org/jenkinsci/plugins/workflow/util/StaplerReferer.java#L61

src/main/java/com/cloudbees/plugins/credentials/ContextInPath.java
```diff
@@ -0,0 +1,46 @@
...
+ * {@link CredentialsSelectHelper.ContextResolver#getToken(ModelObject)} respectively.
+ *
+ * @see CredentialsDescriptor#getCheckMethod(String)
+ * @since 2.1.5
+ */
+@Retention(RUNTIME)
+@Target(PARAMETER)
+@Documented
+@InjectedParameter(ContextInPath.HandlerImpl.class)
+public @interface ContextInPath {
```
stephenc: @jglick you may be interested in this as a less hacky approach to https://github.com/jenkinsci/workflow-step-api-plugin/blob/aabbdca1c3c1f1aacf346b01d95d3be467547195/src/main/java/org/jenkinsci/plugins/workflow/util/StaplerReferer.java#L43

I wonder could we introduce some extension points and make a more generic solution in core?

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -63,22 +88,16 @@ protected CredentialsDescriptor() {
...
+    @Restricted(NoExternalUse.class)
+    @RestrictedSince("2.1.5")
+    public ListBoxModel doFillScopeItems(@ContextInPath ModelObject context) {
         ListBoxModel m = new ListBoxModel();
-        Ancestor ancestor = Stapler.getCurrentRequest().findAncestor(Object.class);
-        while (ancestor != null) {
-            if (ancestor.getObject() instanceof ModelObject) {
-                Set<CredentialsScope> scopes =
-                        CredentialsProvider.lookupScopes((ModelObject) ancestor.getObject());
+                Set<CredentialsScope> scopes = CredentialsProvider.lookupScopes(context);
```
oleg-nenashev: NIT. formatting. Or maybe dead code?

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,339 @@ public String getIconClassName() {
...
+     * classes.
+     * This helper method unwraps those to return the correct context.
+     *
+     * @param context the context (wrapped or unwrapped).
+     * @return the unwrapped context.
+     * @since 2.1.5
+     */
+    @NonNull
+    public static ModelObject unwrapContext(@NonNull ModelObject context) {
+        if (context instanceof CredentialsSelectHelper.WrappedCredentialsStore) {
```
rsandell: same `getStore()` method on all classes. Seems like an interface could clean this up a bit and lend to easier future expansion.

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,339 @@ public String getIconClassName() {
...
+                    //         MMMMMMMMMMMMMMMM      MMMMMMM   MMMMMMMM
+                    //        MMMMMMMMM.MMMMMMM      MMMMMMM MMMMMMMMMM.
+                    //        MMMMMMMMMMMMMMMM?      MMMMMM MMMMMMMMMMM
+                    //        MMMMMMMMMM  .  .       MMMMMIMMMMMMMMMMMM.
+                    //        MMMMMMMMMM               .. :MMMMMMMMMMMM.
+                    //        DMMMMMMMMMM                 MMMMMMMMMMMMM.
+                    //         MMMMMMMMMM.M.              MMMMMMMMMMMM.
+                    //           MMMMMM,                    ....
+                    //
+                    //                   I AM A SAD PANDA
```
oleg-nenashev: 🐜 You should feed your panda better :) But it would be great to have an issue link

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,339 @@ public String getIconClassName() {
...
+     * @param provider the provider.
+     * @param token the token.
+     * @return the context.
+     *
+     * @since 2.1.5
+     */
+    public static ModelObject lookupContext(String provider, String token) {
+        for (CredentialsSelectHelper.ContextResolver r : ExtensionList
+                .lookup(CredentialsSelectHelper.ContextResolver.class)) {
+            if (r.getClass().getName().equals(provider)) {
```
oleg-nenashev: Maybe it should resolve class renaming routines defined for XStream. Looks reasonable for Model Objects

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,339 @@ public String getIconClassName() {
...
+
+    /**
+     * Looks up the context given the provider and token.
+     * @param provider the provider.
+     * @param token the token.
+     * @return the context.
+     *
+     * @since 2.1.5
+     */
+    public static ModelObject lookupContext(String provider, String token) {
```
oleg-nenashev: Undocumented null (no Javadoc/Annotation). Even if it's a common behavior for this method, better to require checks

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
rsandell: :sadpanda: I've had need in the past fo something similar in Stapler as well
https://github.com/jenkinsci/build-failure-analyzer-plugin/blob/master/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/BuildLogIndication.java#L263

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc: Why? This is only ever invoked from the list of `ContextResolver` implementations in the preceding stapler response that then generated this request

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc: shouldn't be future expansion ;-)

But yeah, if I could think of a decent name...

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,340 @@ public String getIconClassName() {
...
+            }
+        }
+        return null;
+
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public FormValidation.CheckMethod getCheckMethod(String fieldName) {
```
oleg-nenashev: Should this hack be restricted (from direct override calls)?

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,340 @@ public String getIconClassName() {
...
+                JellyContext jelly = Functions.getCurrentJellyContext();
+                Object it = jelly.findVariable("it");
+                if (it instanceof CredentialsStore) {
+                    ModelObject context = ((CredentialsStore) it).getContext();
+                    for (CredentialsSelectHelper.ContextResolver r : ExtensionList
+                            .lookup(CredentialsSelectHelper.ContextResolver.class)) {
+                        String token = r.getToken(context);
+                        if (token != null) {
+                            String fillUrl = (String) attributes.get("fillUrl");
+                            if (fillUrl.indexOf('?') != -1) {
```
oleg-nenashev: Check for null?

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc: I think @jglick has an issue raised already but my JIRA-foo is not good at finding it

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,340 @@ public String getIconClassName() {
...
+                                fillUrl = fillUrl + '&';
+                            } else {
+                                fillUrl = fillUrl + '?';
+                            }
+                            attributes.put("fillUrl", fillUrl + "$provider=" + r.getClass().getName() + "&"
```
oleg-nenashev: :bug: Class name may contain "$" Symbols. Needs escaping

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,340 @@ public String getIconClassName() {
...
+                                    + "$token=" + URLEncoder.encode(token, "UTF-8"));
+                        }
+                    }
+
+                }
+            } catch (AssertionError e) {
+                // ignore, we did the best we could
```
oleg-nenashev: Log?

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,340 @@ public String getIconClassName() {
...
+            } catch (UnsupportedEncodingException e) {
+                // ignore, we did the best we could
```
oleg-nenashev: SEVERE log? YourJenkinsIsNotGonnaWorkException

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,340 @@ public String getIconClassName() {
...
+                JellyContext jelly = Functions.getCurrentJellyContext();
+                Object it = jelly.findVariable("it");
+                if (it instanceof CredentialsStore) {
+                    ModelObject context = ((CredentialsStore) it).getContext();
+                    for (CredentialsSelectHelper.ContextResolver r : ExtensionList
+                            .lookup(CredentialsSelectHelper.ContextResolver.class)) {
+                        String token = r.getToken(context);
+                        if (token != null) {
+                            String fillUrl = (String) attributes.get("autoCompleteUrl");
+                            if (fillUrl.indexOf('?') != -1) {
```
oleg-nenashev: Check for null

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc: Found it: https://issues.jenkins-ci.org/browse/JENKINS-19413

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,340 @@ public String getIconClassName() {
...
+                                fillUrl = fillUrl + '&';
+                            } else {
+                                fillUrl = fillUrl + '?';
+                            }
+                            attributes.put("autoCompleteUrl", fillUrl + "$provider=" + r.getClass().getName() + "&"
```
oleg-nenashev: as above

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,340 @@ public String getIconClassName() {
...
+                                    + "$token=" + URLEncoder.encode(token, "UTF-8"));
+                        }
+                    }
+
+                }
+            } catch (AssertionError e) {
+                // ignore, we did the best we could
+            } catch (UnsupportedEncodingException e) {
```
oleg-nenashev: as above

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc: yeah... mine is better I think ;-)

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,340 @@ public String getIconClassName() {
...
+        public EnhancedCheckMethod(Descriptor descriptor, String fieldName) {
+            super(descriptor, fieldName);
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public String toCheckUrl() {
+            String checkUrl = super.toCheckUrl();
```
oleg-nenashev: 🐛 This method in Jenkins may return null. It's gonna cause NPE in the code

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
rsandell: `CredentialStoreOwner`?

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
oleg-nenashev: https://github.com/jenkinsci/jenkins/blob/9fce1ee933eb5276baff977d562fc8e183f1c8d6/core/src/main/java/hudson/util/FormValidation.java#L615

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,343 @@ public String getIconClassName() {
...
+                                    + ".append('$token=" + Functions.jsStringEscape(
+                                    URLEncoder.encode(token, "UTF-8")) + "')"
+                                    + ".toString()";
+                        }
+                    }
+
+                }
+            } catch (AssertionError e) {
+                // ignore, we did the best we could
+            } catch (UnsupportedEncodingException e) {
```
oleg-nenashev: logging?

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc:
```
pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
```

and 

```
sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
                  / "*" / "+" / "," / ";" / "="
```

So $ does not need escaping

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
rsandell: Yes, I didn't write mine myself and IIRC it doesn't work well with views and folders :) But hardly none has complained, probably because it isn't a well known feature.

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc: But it is safer to escape anyway... just pointing out that your stated reason is invalid... there is a valid reason to escape!

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc: no

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc: no

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,343 @@ public String getIconClassName() {
...
+    @Override
+    public void calcFillSettings(String field, Map<String, Object> attributes) {
+        if (attributes.containsKey("fillUrl")) {
+            // the user already provided a custom one, get out of the way
+            super.calcFillSettings(field, attributes);
+            return;
+        }
+        // this is an ugly hack to make the @ContextInPath annotation more failsafe
+        super.calcFillSettings(field, attributes);
+        if (attributes.containsKey("fillUrl")) {
```
rsandell: :bug: Same check as above, so this code will not run.

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc: no

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc: nope... the `calcFillSettings` method will add the attribute.

The idea is to do a no-op if the user supplied the fillUrl, but if the fillUrl was calculated then we add the `$provider` and `$token`

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc: Nope because the ModelObject or the CredentialsProvider is the owner

src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java
```diff
@@ -145,11 +154,16 @@ private static FormValidation checkForDuplicates(String value, ModelObject conte
...
-                            != null) {
+                    Credentials match = CredentialsMatchers
+                            .firstOrNull(store.getCredentials(domain), CredentialsMatchers.withId(value));
+                    if (match != null) {
                         if (storeContext == context) {
                             return FormValidation.error("This ID is already in use");
                         } else {
+                            if (match.getScope() == CredentialsScope.SYSTEM) {
+                                // system scope is not exported to child contexts
+                                continue;
```
oleg-nenashev: 🐛 Should it continue with other credentials in the domain then? firstOrNull skips others from what I see

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
rsandell: Then :ant: A comment mentioning that would be nice ;)

src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java
```diff
@@ -161,7 +175,28 @@ private static FormValidation checkForDuplicates(String value, ModelObject conte
...
+            ModelObject context = store.getContext();
+            for (ContextResolver r : ExtensionList.lookup(ContextResolver.class)) {
+                String token = r.getToken(context);
+                if (token != null) {
+                    return Jenkins.getActiveInstance().getRootUrlFromRequest() + "/descriptorByName/" + clazz.getName()
+                            + "/checkId?provider=" + r.getClass().getName() + "&token="
+                            + URLEncoder.encode(token, "UTF-8");
+                }
+            }
+            return Jenkins.getActiveInstance().getRootUrlFromRequest() + "/descriptorByName/" + clazz.getName()
```
oleg-nenashev: 🐜 class name escaping

src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java
```diff
@@ -161,7 +175,28 @@ private static FormValidation checkForDuplicates(String value, ModelObject conte
...
+         * @return the url of the id check endpoint.
+         * @throws UnsupportedEncodingException if the JVM does not implement the JLS.
+         */
+        public String getCheckIdUrl(CredentialsStore store) throws UnsupportedEncodingException {
+            ModelObject context = store.getContext();
+            for (ContextResolver r : ExtensionList.lookup(ContextResolver.class)) {
+                String token = r.getToken(context);
+                if (token != null) {
+                    return Jenkins.getActiveInstance().getRootUrlFromRequest() + "/descriptorByName/" + clazz.getName()
+                            + "/checkId?provider=" + r.getClass().getName() + "&token="
```
oleg-nenashev: 🐜 class name escaping

src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java
```diff
@@ -161,7 +175,28 @@ private static FormValidation checkForDuplicates(String value, ModelObject conte
...
+                            + URLEncoder.encode(token, "UTF-8");
+                }
+            }
+            return Jenkins.getActiveInstance().getRootUrlFromRequest() + "/descriptorByName/" + clazz.getName()
+                    + "/checkId?provider=null&token=null";
+        }
+
+        public final FormValidation doCheckId(@ContextInPath ModelObject context, @QueryParameter String value) {
```
oleg-nenashev: Restricted?

src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java
```diff
@@ -172,6 +207,7 @@ public final FormValidation doCheckId(@QueryParameter String value, @AncestorInP
             if (problem != null) {
                 return problem;
             }
+            CredentialsStoreAction a;
```
oleg-nenashev: dead code

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -192,4 +257,356 @@ public String getIconClassName() {
...
+    @Override
+    public void calcAutoCompleteSettings(String field, Map<String, Object> attributes) {
+        if (attributes.containsKey("autoCompleteUrl")) {
+            // the user already provided a custom one, get out of the way
+            super.calcAutoCompleteSettings(field, attributes);
+            return;
+        }
+        // this is an ugly hack to make the @ContextInPath annotation more failsafe
+        super.calcAutoCompleteSettings(field, attributes);
+        if (attributes.containsKey("autoCompleteUrl")) {
```
rsandell: So, seems like only a one string difference between this and the code for fillUrl above. Reducing the "hack" to a one method call would help with scrubbing it in the future.

rsandell:
:bee:

oleg-nenashev:
I keep 🐛 for checkForDuplicates() domain handling and className escaping, but the most of the issues have been addressed

src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java
stephenc: talk to core! (but I'll switch to sharing core's bug so I can share in core's fix)

src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
stephenc: you mean like `// the user already provided a custom one, get out of the way`

oleg-nenashev:
"Let's all share core's lack of escaping of descriptor ids" looks good to me since there is no flaws in the known OSS code

stephenc:
@oleg-nenashev are you removing your bug then?

oleg-nenashev:
Removed bug, need to perform another iteration before adding a bee

rsandell:
:bee:

src/main/resources/com/cloudbees/plugins/credentials/CredentialsSelectHelper/WrappedCredentialsStore/credential.jelly
```diff
@@ -24,21 +24,24 @@
...
-        lazy="descriptor,it">
-        <l:ajax>
-          <j:set var="instance" value="${current.descriptor==descriptor ? current : null}" />
-          <st:include from="${descriptor}" page="${descriptor.configPage}" />
-        </l:ajax>
-      </f:dropdownListBlock>
-    </j:forEach>
-  </f:dropdownList>
+  <j:local>
+    <j:set var="it" value="${it.store}"/>
```
jglick: Do you really need `j:local` here? AFAIK `it` would have been scoped only to this include anyway.

jglick:
The JENKINS-19413 workaround does look like a modest improvement over the existing implementations (there are at least two copies lying around that I know of); these should I think be moved into a library plugin for eventual inclusion in core (or Stapler itself). Obviously the better approach would be to directly reuse the URL-binding logic in `Stapler.tryInvoke`, which would seem to require an implementation in Stapler core itself.

The rest of the PR I do not really understand. Something about mismatched `it`s. Do not see anything obviously wrong, other than the usual nightmare of making form validation work in exotic circumstances.

src/main/resources/com/cloudbees/plugins/credentials/CredentialsSelectHelper/WrappedCredentialsStore/credential.jelly
jsoref: I'm curious about this `j:local` -- it bleeds into the html that web browsers see. That isn't inherently bad, but in general one shouldn't bleed xml namespaced tags to html web browser.

stephenc: It shouldn’t, assuming the xml namespace is correctly defined

slide: I don't see a <local> tag in the Jelly documentation

stephenc: Ahh should be `scope` http://commons.apache.org/proper/commons-jelly/tags.html#core:scope Do you want to fix it?

PR #68, opened 2016-09-08
[JENKINS-38048] Relax nullability on Item/ItemGroup context parameters

jglick:
Used implicitly by https://github.com/jenkinsci/git-plugin/pull/437 and https://github.com/jenkinsci/subversion-plugin/pull/169.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/common/AbstractIdCredentialsListBoxModel.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/common/AbstractIdCredentialsListBoxModel.java
@@ -29,6 +29,7 @@ import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.domains.DomainRequirement;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
 import hudson.model.Descriptor;
 import hudson.model.Item;
 import hudson.model.ItemGroup;
@@ -292,7 +293,7 @@ public abstract class AbstractIdCredentialsListBoxModel<T extends AbstractIdCred
      * @see CredentialsProvider#listCredentials(Class, Item, Authentication, List, CredentialsMatcher)
      * @since 2.1.0
      */
-    public AbstractIdCredentialsListBoxModel<T, C> include(@NonNull Item context, @NonNull Class<? extends C> type) {
+    public AbstractIdCredentialsListBoxModel<T, C> include(@Nullable Item context, @NonNull Class<? extends C> type) {
         return include(context, type, Collections.<DomainRequirement>emptyList());
     }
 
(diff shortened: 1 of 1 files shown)


jglick:
Can you explain why?

These newly introduced methods take a `@Nonnull` parameter, then do nothing with it except delegate to longstanding methods accepting a `@Nullable` parameter (with an IMO sensible meaning of looking only in global scope).

stephenc:
Well if we allow `null` then we will get people passing `null` and then the `null` needs to be typecast... plus I now believe it was a mistake to allow the previous APIs to accept `null` rather than forcing people to be explicit.

I am open to debate on this subject. Just my initial push-back

jglick:
> forcing people to be explicit

Well instead of

```java
return new StandardListBoxModel()
        .includeEmptyValue()
        .includeMatchingAs(
                project instanceof Queue.Task
                        ? Tasks.getAuthenticationOf((Queue.Task) project)
                        : ACL.SYSTEM,
                project,
                StandardUsernameCredentials.class,
                GitURIRequirementsBuilder.fromUri(url).build(),
                GitClient.CREDENTIALS_MATCHER)
        .includeCurrentValue(credentialsId);
```

you could inline the implementation of `includeMatchingAs` to satisfy the _existing_ nullability signatures:

```java
StandardListBoxModel model = new StandardListBoxModel().includeEmptyValue();
model.addMissing(CredentialsProvider.listCredentials(
    StandardUsernameCredentials.class,
    project,
    project instanceof Queue.Task
        ? Tasks.getAuthenticationOf((Queue.Task) project)
        : ACL.SYSTEM,
    GitURIRequirementsBuilder.fromUri(url).build(),
    GitClient.CREDENTIALS_MATCHER)));
return model.includeCurrentValue(credentialsId);
```

and then to satisfy proposed _stricter_ nullability you would have to say

```java
StandardListBoxModel model = new StandardListBoxModel().includeEmptyValue();
Authentication authentication = project instanceof Queue.Task
        ? Tasks.getAuthenticationOf((Queue.Task) project)
        : ACL.SYSTEM;
List<DomainRequirement> domainRequirements = GitURIRequirementsBuilder.fromUri(url).build();
CredentialsMatcher matcher = GitClient.CREDENTIALS_MATCHER);
model.addMissing(project != null ?
    CredentialsProvider.listCredentials(
        StandardUsernameCredentials.class,
        project,
        authentication,
        domainRequirements,
        matcher) :
    CredentialsProvider.listCredentials(
        StandardUsernameCredentials.class,
        Jenkins.getInstance(),
        authentication,
        domainRequirements,
        matcher));
return model.includeCurrentValue(credentialsId);
```

which does not feel like an improvement to me, much as I also like explicit APIs.

stephenc:
Compromise suggestion... make the `ItemGroup` variants can be `@Nullable` the `Item` cannot... I am open to vice-versa if you feel that makes more sense...

jglick:
> make the `ItemGroup` variants can be `@Nullable` the `Item` cannot

Well the motivating case (as shown above) involves a null `Item`.

> I am open to vice-versa

Why would it make sense to allow null in some calls but not others? That sort of inconsistency is the whole point of this PR.

jglick:
Conclusion with @stephenc:

* for the short term, let `Item` overloads be `@Nullable`, but `ItemGroup` overloads remains `@Nonnull`
* file an issue to provide a better system:
   * new methods (not overloads) taking `@Nonnull ModelObject`
   * for non-`Item`/`ItemGroup`, currently just fall back to `Jenkins`, but later could handle `User`, `Node`, etc.
   * introduce a helper method for `Authentication` based on item with possible defaults:
```java
project instanceof Queue.Task ? Tasks.getAuthenticationOf((Queue.Task) project): ACL.SYSTEM
```

jglick:
Made the requested short-term change and filed [JENKINS-39324](https://issues.jenkins-ci.org/browse/JENKINS-39324) for the rest.

PR #70, opened 2016-10-31
[JENKINS-20139] Inaccurate/incomplete Javadoc for `track` overloads

jglick:
Amends #58. Noticed from https://github.com/jenkinsci/credentials-binding-plugin/pull/27.

@reviewbybees esp. @stephenc

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
@@ -1445,7 +1445,7 @@ public abstract class CredentialsProvider extends Descriptor<CredentialsProvider
 
     /**
      * Track the usage of credentials in a specific node.
-     *
+     * Would be used for example when launching an agent.
      * @param node        the node to tag the fingerprint
      * @param credentials the credentials to fingerprint.
      * @param <C>         the credentials type.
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -1444,7 +1444,7 @@ public static Fingerprint getOrCreateFingerprintOf(@NonNull Credentials c) throw
     }
 
     /**
-     * Track the usage of credentials in a specific build
+     * Track the usage of credentials in a specific node.
```
stephenc: Include the comment on launching an agent

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -1461,7 +1461,7 @@ public static Fingerprint getOrCreateFingerprintOf(@NonNull Credentials c) throw
...
```
stephenc: Include the comment on launching an agent

stephenc: E.g. Launching an agent

PR #71, opened 2016-11-04
[JENKINS-39516] Test failing against jenkins core > 1.626

raul-arabaolaza:
[JENKINS-39516](https://issues.jenkins-ci.org/browse/JENKINS-39516)

@reviewbybees specially @stephenc


raul-arabaolaza:
@stephenc Changed to check for action type instead of returning the first

PR #73, opened 2016-11-16
[FIXED JENKINS-39317] Provide a mechanism for forcing a save of all credential store

stephenc:
- This method will only be available via Groovy scripting

See [JENKINS-39317](https://issues.jenkins-ci.org/browse/JENKINS-39317)

@reviewbybees


craigbarrau:
Looks like a useful feature @stephenc 
How would one trigger the encryption via groovy on a credentials.xml containing plaintext?

stephenc:
`com.cloudbees.plugins.credentials.CredentialsProvider.saveAll()`

craigbarrau:
Thanks!

PR #78, opened 2017-01-26
[FIXED JENKINS-41478] Add button should be enabled if any stores are enabled

stephenc:
- Also disabled stores should be disabled in the menu

See [JENKINS-41478](https://issues.jenkins-ci.org/browse/JENKINS-41478)

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
@@ -180,6 +180,37 @@ public class CredentialsSelectHelper extends Descriptor<CredentialsSelectHelper>
     }
 
     /**
+     * Checks if the current user has permission to create a credential.
+     *
+     * @param context     the context.
+     * @param includeUser whether they can use their own credentials store.
+     * @return {@code true} if they can create a permission.
+     * @since FIXME
+     */
+    @Restricted(NoExternalUse.class)
+    @SuppressWarnings("unused") // used via jelly
+    public boolean hasCreatePermission(ModelObject context, boolean includeUser) {
+        if (includeUser) {
+            User current = User.current();
+            if (current != null && current.hasPermission(CREATE)) {
+                return true;
+            }
+        }
+        if (context == null) {
+            StaplerRequest request = Stapler.getCurrentRequest();
+            if (request != null) {
+                context = request.findAncestorObject(ModelObject.class);
+            }
+        }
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
```diff
@@ -180,6 +180,37 @@ public ModelObject resolveContext(Object context) {
...
+        if (includeUser) {
+            User current = User.current();
+            if (current != null && current.hasPermission(CREATE)) {
+                return true;
+            }
+        }
+        if (context == null) {
+            StaplerRequest request = Stapler.getCurrentRequest();
+            if (request != null) {
+                context = request.findAncestorObject(ModelObject.class);
```
jglick: And if that is null also, do we gracefully return false, or does `lookupStores` give an NPE?

stephenc:
screen shot 2017-01-26 at 16 21 25

Here is the Add button now enabled even though the current user does not have create permission at the root element, but does in the folder

Note that the folder is selectable while the root store is disabled as it is read-only to the current user

src/main/resources/lib/credentials/select.jelly
```diff
@@ -87,13 +87,24 @@
...
+                  <li class="yuimenuitem" data-url="${request.contextPath}/${storeItem.url}/dialog">
+                    <span class="yuimenuitemlabel" tooltip="${storeItem.description}">
+                      <l:icon class="${storeItem.iconClassName} icon-sm"/>
+                      ${storeItem.displayName}
+                    </span>
+                  </li>
+                </j:when>
+                <j:otherwise>
+                  <li class="yuimenuitem" disabled="disabled">
+                    <span class="yuimenuitemlabel" tooltip="${storeItem.description}" disabled="disabled">
```
jglick: Would make for a smaller diff and simpler code to introduce a variable for `storeItem.enabled ? null : 'disabled'` and use it on both the `li` and the `span`.

src/main/resources/lib/credentials/select/select.js
```diff
@@ -200,6 +200,17 @@ Behaviour.specify("BUTTON.credentials-add-menu", 'credentials-select', -99, func
             }
             window.credentials.add(item.srcElement.getAttribute('data-url'));
         });
+        // YUI menu will not parse disabled when using DIV-LI only when using SELECT-OPTION
```
jglick: Oof. Will we ever be able to replace YUI without having to go all the way to Blue Ocean?

PR #80, opened 2017-01-30
[JENKINS-41512] Use @media query to scale select max-width

stephenc:
See [JENKINS-41512](https://issues.jenkins-ci.org/browse/JENKINS-41512)

@reviewbybees 

ezgif com-optimize


stephenc:
@gusreiber has reviewed and could not suggest a better fix

PR #82, opened 2017-02-11
[JENKINS-41946] Fix Stapler "IllegalArgumentException: Unable to convert to class com.cloudbees.plugins.credentials.SecretBytes"

cyrille-leclerc:
[JENKINS-41946] Fix Stapler "IllegalArgumentException: Unable to convert to class com.cloudbees.plugins.credentials.SecretBytes".

Add `com.cloudbees.plugins.credentials.SecretBytes$StaplerConverterImpl`.

@stephenc can we remove the static manual registration of the `SecretBytes` converter in `org.kohsuke.stapler.Stapler#CONVERT_UTILS` thanks to https://github.com/jenkinsci/credentials-plugin/blob/credentials-2.1.11/src/main/java/com/cloudbees/plugins/credentials/SecretBytes.java#L358 ?

PR #83, opened 2017-02-14
[JENKINS-41952] mismatch on the encoding of the keystore: sometimes as `Secret(base64(keystore))` and sometimes as `new SecretBytes(keystore).toString()`

cyrille-leclerc:
[JENKINS-41952] mismatch on the encoding of the keystore: sometimes as `Secret(base64(keystore))` and sometimes as `new SecretBytes(keystore).toString()`

ℹ️  PR started from #82

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
@@ -603,8 +603,7 @@ public class CertificateCredentialsImpl extends BaseStandardCredentials implemen
             public static byte[] toByteArray(@Nullable Secret secret) {
                 if (secret != null) {
                     try {
-                        String plainText = secret.getPlainText();
-                        return Base64.decode(plainText.toCharArray());
+                        return Base64.decode(secret.getPlainText().toCharArray());
                     } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


cyrille-leclerc:
@stephenc I did not dare to cleanup the code for the moment, there are 2 paths of validation, one for each encryption of the keystore.

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
```diff
@@ -640,7 +648,30 @@ public FormValidation doCheckUploadedKeystore(@QueryParameter String value,
                 if (StringUtils.isBlank(value)) {
                     return FormValidation.error(Messages.CertificateCredentialsImpl_NoCertificateUploaded());
                 }
-                return validateCertificateKeystore("PKCS12", toByteArray(Secret.fromString(value)), password);
+                byte[] keystoreBytes;
+
+                // JENKINS-41946: value can have 2 different formats
+                if (value.startsWith("{") && value.endsWith("}")) {
```
stephenc: this will not be sufficient as on latest cores all secrets, even ones from `Secret` will start and end with `{` and `}`

I think we need to ensure a consistent usage instead rather than trying to dynamically detect

PR #86, opened 2017-06-01
[FIXED JENKINS-44588] Add "credentials" symbol to parameter

abayer:
Allows use of the credentials parameter type in Declarative.

cc @reviewbybees and @vivek


vivek:
LGTM 🐝

abayer:
@stephenc Ping? I don't feel comfortable merging here without your go-ahead.

PR #92, opened 2017-09-14
[WIP] Allow non-modifiable credentials stores

hoshsadiq:
This pull request refactors non-modifiable domains and also allows for non-modifiable credentials stores. This is specifically when you don't want the credentials store to be editable through Jenkins but you still need a third party credentials store. My use case is a plugin I'm working on (https://github.com/hoshsadiq/jenkins-kubernetes-secrets-credentials), where the credentials are written through Kubernetes' kubectl, and they should be readable through Jenkins.


hoshsadiq:
Discussed with @stephenc, this PR is too risky, will raise a more simpler one.

PR #93, opened 2017-10-06
[JENKINS-46859] Allow read-only credentials providers/stores

hoshsadiq:
Simply allows ReadOnly credentials provider to be implemented more simpler. Closes [JENKINS-46859](https://issues.jenkins-ci.org/browse/JENKINS-46859).

This should be revisited in stages, however, as it stands it's to do a better implementation (an attempt is PR #92). After discussing with @stephenc, this seemed a simpler and less risky solution.

Docs yet to be updated.

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsStore.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsStore.java
@@ -261,7 +261,7 @@ public abstract class CredentialsStore implements AccessControlled, Saveable {
      * <li>{@link #updateDomain(Domain, Domain)} </li>
      * </ul>
      *
-     * @return {@code true} if {@link #addDomain(Domain, List)}
+     * @return {@code true} iff {@link #addDomain(Domain, List)}
      * {@link #addDomain(Domain, Credentials...)}, {@link #removeDomain(Domain)}
      * and {@link #updateDomain(Domain, Domain)} are expected to work
      */
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsStore.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsStore.java
@@ -81,11 +81,6 @@ public abstract class CredentialsStore implements AccessControlled, Saveable {
     private transient Boolean domainsModifiable;
 
     /**
-     * Cache for {@link #isCredentialsModifiable()}.
-     */
-    private transient Boolean credentialsModifiable;
-
-    /**
      * Constructor for use when the {@link CredentialsStore} is not an inner class of its {@link CredentialsProvider}.
      *
      * @param providerClass the {@link CredentialsProvider} class.
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsStore.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsStore.java
@@ -81,6 +81,11 @@ public abstract class CredentialsStore implements AccessControlled, Saveable {
     private transient Boolean domainsModifiable;
 
     /**
+     * Cache for {@link #isCredentialsModifiable()}.
+     */
+    private transient Boolean credentialsModifiable;
+
+    /**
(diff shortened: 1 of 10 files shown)


stephenc:
I haven’t reviewed yet, but the docs will need to be updated in tandem, so please include the normative docs (ie the ones in this repo) update as part of this PR

src/main/java/com/cloudbees/plugins/credentials/CredentialsStore.java
```diff
@@ -256,7 +261,7 @@ public Domain getDomainByName(@CheckForNull String name) {
      * <li>{@link #updateDomain(Domain, Domain)} </li>
      * </ul>
      *
-     * @return {@code true} iff {@link #addDomain(Domain, List)}
+     * @return {@code true} if {@link #addDomain(Domain, List)}
```
stephenc: “IFF” is a common short-hand for “If and Only If” not a typo ;-)

hoshsadiq: My bad! I was not aware of this. I'll change it back 👍

hoshsadiq:
@stephenc, I think it's ready! Let me know if you think more is needed

src/main/java/com/cloudbees/plugins/credentials/CredentialsStore.java
```diff
@@ -81,6 +81,11 @@
     private transient Boolean domainsModifiable;
 
     /**
+     * Cache for {@link #isCredentialsModifiable()}.
+     */
+    private transient Boolean credentialsModifiable;
```
stephenc: No need for the cache. The method itself is `final` and non complex

src/main/java/com/cloudbees/plugins/credentials/ReadOnlyCredentialsStore.java
```diff
@@ -0,0 +1,124 @@
...
+import java.io.IOException;
+
+/**
+ * A read only store of {@link Credentials}. Each {@link ReadOnlyCredentialsStore} is associated with one and only one
+ * {@link CredentialsProvider} though a {@link CredentialsProvider} may provide multiple {@link ReadOnlyCredentialsStore}s
+ * (for example a folder scoped {@link CredentialsProvider} may provide a {@link ReadOnlyCredentialsStore} for each folder
+ * or a user scoped {@link CredentialsProvider} may provide a {@link ReadOnlyCredentialsStore} for each user).
+ *
+ * @author Hosh Sadiq
+ * @since 2.1.17
```
stephenc: `@since TODO` and i’ll s&r on merge

src/main/java/com/cloudbees/plugins/credentials/ReadOnlyCredentialsStore.java
```diff
@@ -0,0 +1,124 @@
...
+ */
+public abstract class ReadOnlyCredentialsStore extends CredentialsStore {
+    /**
+     * The {@link CredentialsProvider} class.
+     *
+     * @since 2.0
```
stephenc: No need for `@since` Within a class where the classes’s since is the same

src/main/java/com/cloudbees/plugins/credentials/ReadOnlyCredentialsStore.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    private final Class<? extends CredentialsProvider> providerClass;
+
+    public ReadOnlyCredentialsStore(Class<? extends CredentialsProvider> providerClass) {
+        this.providerClass = providerClass;
+    }
+
+    public ReadOnlyCredentialsStore() {
+        // now let's infer our provider, Jesse will not like this evil
```
stephenc: Do you really need all this? Is it not inherited?

hoshsadiq: Can't remember the reason, think it was because of previous attempt at the implementation. Removed them now.

PR #95, opened 2017-11-06
Find Credential Parameters approved by a user in a pipeline input step (if any)

b-dean:
Fix for JENKINS-47699

Related to jenkinsci/pipeline-input-step-plugin#19

I'm sure there are going to be complaints about how I did this, particularly in the `CredentialsProvider#getInputParameters` method. I asked on IRC for advice on how to do something like this but didn't get anything of substance. I'm sure there's a "right way" to do this sort of thing and I'm sure this isn't it. Sorry in advance :slightly_smiling_face:


jglick:
This is certainly not the right way to do it. I am not sure offhand what the right way is. Probably requires @stephenc to advise on design.

jvz:
Closing in favor of an alternative approach in progress.

Wadeck:
FTR closed for #119

PR #96, opened 2017-12-01
Updates

jglick:
For the same reason as https://github.com/jenkinsci/cloudbees-folder-plugin/pull/116, but for `SystemCredentialsProviderTest`.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.32</version>
+    <version>3.1</version>
     <relativePath />
   </parent>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -65,6 +65,8 @@
   </scm>
 
   <properties>
+    <jenkins.version>1.625.3</jenkins.version>
+    <java.level>7</java.level>
```
jglick: Almost forgot these! Fortunately https://github.com/jenkinsci/plugin-pom/pull/88 kicked in.

jglick:
@reviewbybees done

pom.xml
stephenc: [Will be going](https://groups.google.com/d/msg/jenkinsci-dev/1xqwDRck-zk/0Fi9Jw82AAAJ) to `2.60.3` and `8` but we can merge this ahead of the bump

PR #97, opened 2018-01-31
[JENKINS-49235] Prune fingerprints of non existent node credentials

fraz3alpha:
[JENKINS-49235](https://issues.jenkins-ci.org/browse/JENKINS-49235)

  - Add a test to check that fingerprints aren't stored for nodes that aren't known
    - Checks that storing a fingerprint for a node that Jenkins doesn't know
      about is not saved
    - Checks that adding a fingerprint for a node that Jenkins does know about
      is saved
    - Checks that the fingerprint is removed when calling track() after
      the node has been removed
    - Check that tracking a credential on the same node again doesn't
      result in two entries

  - Add functionality for the above. When iterating the facets of the
    fingerprint any NodeCredentialsFingerprintFacet elements are checked
    in the iterator and removed if necessary. An additional check has been
    added around the eventual addition to skip the case where it doesn't
    exist at that point

cf. https://github.com/jenkinsci/ssh-slaves-plugin/pull/35

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
@@ -1526,14 +1526,13 @@ public abstract class CredentialsProvider extends Descriptor<CredentialsProvider
                     try {
                         Collection<FingerprintFacet> facets = fingerprint.getFacets();
                         // purge any old facets
-                        // Current
                         long start = timestamp;
                         for (Iterator<FingerprintFacet> iterator = facets.iterator(); iterator.hasNext(); ) {
                             FingerprintFacet f = iterator.next();
                             // For all the node-tracking credentials, check to see if we can remove
                             // older instances of these credential fingerprints, or from nodes which no longer exist
                             if (f instanceof NodeCredentialsFingerprintFacet) {
-                                // Remove older instance
+                                // Remove older instances
                                 if (StringUtils.equals(nodeName, ((NodeCredentialsFingerprintFacet) f).getNodeName())) {
                                     start = Math.min(start, f.getTimestamp());
                                     iterator.remove();
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
@@ -1771,3 +1771,4 @@ public abstract class CredentialsProvider extends Descriptor<CredentialsProvider
         }
     }
 }
+
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
@@ -1515,20 +1515,38 @@ public abstract class CredentialsProvider extends Descriptor<CredentialsProvider
                 try {
                     Fingerprint fingerprint = getOrCreateFingerprintOf(c);
                     BulkChange change = new BulkChange(fingerprint);
+
+                    // Create a list of all current node names, the
+                    // credential will only be fingerprinted if it is one of these
+                    Set<String> jenkinsNodeNames = new HashSet<String>();
+                    for (Node n: Jenkins.getInstance().getNodes()) {
+                        jenkinsNodeNames.add(n.getNodeName());
+                    }
+
                     try {
                         Collection<FingerprintFacet> facets = fingerprint.getFacets();
                         // purge any old facets
+                        // Current
                         long start = timestamp;
                         for (Iterator<FingerprintFacet> iterator = facets.iterator(); iterator.hasNext(); ) {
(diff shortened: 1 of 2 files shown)


src/test/java/com/cloudbees/plugins/credentials/CredentialsProviderTest.java
```diff
@@ -59,7 +68,7 @@
 
     @Rule
     public JenkinsRule r = new JenkinsRule();
-
```
stephenc: Traditionally we fix whitespace and other formatting changes in a dedicated PR. Could you move those changes to a different PR please

fraz3alpha: Thanks for the feedback. I think this was because Atom trimmed blank lines when I edited the file. Are you happy if I push another commit to revery these, or would you like me to squash all the ones I've made so far and re-push to my branch?

fraz3alpha:
@stephenc @jglick I have removed the errant spaces as requested - there should be no extra formatting changes now.

fraz3alpha:
@stephenc @jglick Can you help me work out what needs to be done to progress this further please? I have reverted the erroneously added formatting changes as per the request of @stephenc , but I don't know what to do next - I see that the pr-merge job failed, but it is not clear to me why

docwhat:
> Possible null pointer dereference in `com.cloudbees.plugins.credentials.CredentialsProvider.trackAll(Node, List)` due to return value of called method `[com.cloudbees.plugins.credentials.CredentialsProvider, com.cloudbees.plugins.credentials.CredentialsProvider]` Dereferenced at `CredentialsProvider.java:[line 1522]` Known `null` at `CredentialsProvider.java:[line 1522]` `NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE`

See https://ci.jenkins.io/job/Plugins/job/credentials-plugin/job/PR-97/2/console for the old style.

The actual problem was in the Windows part, but since Windows died it sent an abort to the linux side which is why both are red in the blueocean view.

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -1515,20 +1515,37 @@ public static Fingerprint getOrCreateFingerprintOf(@NonNull Credentials c) throw
                 try {
                     Fingerprint fingerprint = getOrCreateFingerprintOf(c);
                     BulkChange change = new BulkChange(fingerprint);
+
+                    // Create a list of all current node names, the
+                    // credential will only be fingerprinted if it is one of these
+                    Set<String> jenkinsNodeNames = new HashSet<String>();
+                    for (Node n: Jenkins.getInstance().getNodes()) {
```
dwnusbaum: I think the CI failure is because `Jenkins#getInstance` is [nullable](https://github.com/jenkinsci/jenkins/blob/2223dc7b547b5960d323ccc87a5cd62773dee988/core/src/main/java/jenkins/model/Jenkins.java#L711) in the version of Jenkins that the plugin is built against (1.625.3), so FindBugs doesn't like the chained `Jenkins.getInstance().getNodes()` call.

If this code is never intended to run while Jenkins is starting or shutting down then you could change it to `Jenkins.getActiveInstance` which will throw `IllegalStateException` if the instance is null, otherwise you could null check the result of `Jenkins.getInstance()` manually and do whatever makes sense in that case.

dwnusbaum: See https://github.com/fraz3alpha/credentials-plugin/pull/1.

fraz3alpha:
Thanks for your PR to fix the build error @dwnusbaum  - the latest build has passed.

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -1515,20 +1515,38 @@ public static Fingerprint getOrCreateFingerprintOf(@NonNull Credentials c) throw
                 try {
                     Fingerprint fingerprint = getOrCreateFingerprintOf(c);
                     BulkChange change = new BulkChange(fingerprint);
+
+                    // Create a list of all current node names, the
+                    // credential will only be fingerprinted if it is one of these
```
stephenc: So in doing this we have not tracking of when a cloud uses credentials

docwhat:
Can we track against the cloud instead of the agent?

dwnusbaum:
In #103 I've added a Jira ticket to fix ephemeral node tracking and referenced it in the code with a TODO comment as requested.

fraz3alpha:
Thank you @dwnusbaum for picking this up and creating a new PR - can one of the reviewers @stephenc, (@jglick ?) take a look at PR #103 and see if that is now acceptable please?, many thanks

jvz:
Superseded by #103.

PR #101, opened 2018-06-26
[JENKINS-52175] Add a CLI command to list all credentials in a store …

amuniz:
…in XML format

This adds a new command `list-credentials-as-xml` which can be used to list credentials in a store and import them to it (using the `--import` option). The output of the command can be used as input as is, the only needed change is to set the actual `Secret`s which are redacted in the output.

@reviewbybees


src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
```diff
@@ -0,0 +1,62 @@
...
+@Extension
+public class ListCredentialsAsXmlCommand extends BaseCredentialsCLICommand {
+
+    @Argument(metaVar = "STORE", usage = "Store ID. For example the Jenkins credentials global store would " +
+            "be identified by \"system::system::jenkins\"",
+            required = true)
+    private CredentialsStore store;
+
+    @Option(name = "--import", usage = "Read credentials as XML from the standard input and create them in the given store." +
+            "Input format is the same as the output generated by this command without enabling this option.")
```
mikecirioli: This sounds a little odd to me - maybe something like `same as the output of 'ListCredentials' command`?

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
```diff
@@ -0,0 +1,62 @@
...
+    protected int run() throws Exception {
+        if (doImport) {
+            store.checkPermission(CredentialsProvider.UPDATE);
+            List<DomainCredentials> replacement = (List<DomainCredentials>) Items.XSTREAM.unmarshal(safeXmlStreamReader(stdin));
+            for (DomainCredentials domain : replacement) {
+                for (Credentials credentials : domain.getCredentials()) {
+                    store.addDomain(domain.getDomain(), credentials);
+                }
+            }
+            return 0;
```
jtnord: this should print (output) the number of secrets imported...

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
```diff
@@ -0,0 +1,62 @@
...
+        } else {
+            store.checkPermission(CredentialsProvider.VIEW);
+
+            List<DomainCredentials> existing = new ArrayList<>();
+            List<Domain> domains = store.getDomains();
+            for (Domain domain : domains) {
+                existing.add(new DomainCredentials(domain, store.getCredentials(domain)));
+            }
+            CredentialsStoreAction.SECRETS_REDACTED.toXML(existing, new OutputStreamWriter(stdout, "UTF-8"));
```
jtnord: this should print (output) the number of secrets exported - so you can verify that all are imported correctly later on.

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
amuniz: Not, the `list-credentials` command output is not valid as input for this new command.

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    protected int run() throws Exception {
+        if (doImport) {
+            store.checkPermission(CredentialsProvider.UPDATE);
+            List<DomainCredentials> replacement = (List<DomainCredentials>) Items.XSTREAM.unmarshal(safeXmlStreamReader(stdin));
+            for (DomainCredentials domain : replacement) {
+                for (Credentials credentials : domain.getCredentials()) {
+                    store.addDomain(domain.getDomain(), credentials);
```
jtnord: print the id of the secret imported. (because if you are missing a plugin that defines a credential type it will be silently swallowed.

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
mikecirioli: sorry, i meant `list-credentials-as-xml`

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public String getShortDescription() {
+        return "Export and Import credentials as XML";
+    }
+
+    @Override
+    protected int run() throws Exception {
+        if (doImport) {
+            store.checkPermission(CredentialsProvider.UPDATE);
```
jtnord: should have `CREATE`, `UPDATE` (depending on my review comment below) and `MANAGE_DOMAINS` IIUC

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
```diff
@@ -0,0 +1,62 @@
...
+            List<DomainCredentials> replacement = (List<DomainCredentials>) Items.XSTREAM.unmarshal(safeXmlStreamReader(stdin));
+            for (DomainCredentials domain : replacement) {
+                for (Credentials credentials : domain.getCredentials()) {
+                    store.addDomain(domain.getDomain(), credentials);
+                }
+            }
+            return 0;
+        } else {
+            store.checkPermission(CredentialsProvider.VIEW);
```
jtnord: I think this is a little week.  Can you not view a permission without being able to use it?

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.io.OutputStreamWriter;
+import java.util.ArrayList;
+import java.util.List;
+
+@Extension
+public class ListCredentialsAsXmlCommand extends BaseCredentialsCLICommand {
+
+    @Argument(metaVar = "STORE", usage = "Store ID. For example the Jenkins credentials global store would " +
+            "be identified by \"system::system::jenkins\"",
```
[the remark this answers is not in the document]
amuniz: There is already a `@CLIResolver` somewhere in the code.

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
```diff
@@ -0,0 +1,62 @@
...
+    private CredentialsStore store;
+
+    @Option(name = "--import", usage = "Read credentials as XML from the standard input and create them in the given store." +
+            "Input format is the same as the output generated by this command without enabling this option.")
+    private boolean doImport;
+
+
+    @Override
+    public String getShortDescription() {
+        return "Export and Import credentials as XML";
```
mikecirioli: consider adding this to the short description:
> The output of the command can be used as input as is, the only needed change is to set the actual Secrets which are redacted in the output.

and removing the bit on line 29

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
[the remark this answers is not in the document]
amuniz: Existing logic of `store.addDomain` is: do nothing if the credentials already exists in the domain, add it otherwise. If the domain does not exist, then create the domain and add all credentials.

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
amuniz: Difficult to implement without bigger changes in the API. `CredentialsStore.addDomain()` returns `true` if the store was modified. The implementation if `addDomain` would need to be modified or copied&pasted here to get the information.

I don't think it worth the change.

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
amuniz: Then a wrapper object is needed (downsides of exporting the model 😃). My intention is keep the change as short as possible.

The information is in the XML, I can image a very simple Xpath to count nodes.

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
amuniz: Hmm, I think the import will fail before reaching this point. `Items.XSTREAM.unmarshal` will throw a class not found exception when trying to deserialize the XML.

amuniz:
Comments fixed @mikecirioli @jtnord 
Thanks for the review.

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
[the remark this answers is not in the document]
jtnord: I meant as a a user of this cli :-)  The CLIResolver is code for a developer.
ie does it need an extra example for a folder (and not sure what the 2 systems are for) - I wouldn;t be able to guess from reading the `--help` output!

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
jtnord: I forgot there is no possibility to write to a file these days so that stdout can contain messages :-/

jtnord:
can we get a clearer help on the --store for the CLI.  As a user I would have no idea what to put hre except litterally what is in the example, but if I want to export credentials from a folder I would be clueless.

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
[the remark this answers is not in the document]
amuniz: Totally agree. I had to deep dive in the code to guess what the `system::system::jenkins` string means (and I'm still not sure). This is the way *all* CLI commands in this plugin work. It is not my goal to change that now, but I can add an example for a folder, sure.

src/test/java/com/cloudbees/plugins/credentials/cli/CLICommandsTest.java
```diff
@@ -306,6 +309,47 @@ public void deleteSmokes() throws Exception {
         assertThat(store.getDomainByName("smokes"), nullValue());
     }
 
+    @Test
+    public void listCredentialsAsXML() throws IOException {
+        Domain smokes = new Domain("smokes", "smoke test domain",
+                Collections.<DomainSpecification>singletonList(new HostnameSpecification("smokes.example.com", null)));
+        UsernamePasswordCredentialsImpl smokey =
+                new UsernamePasswordCredentialsImpl(CredentialsScope.GLOBAL, "smokes-id", "smoke testing", "smoke",
```
jtnord: can you add a credential at system scope to the test?

src/main/java/com/cloudbees/plugins/credentials/cli/ListCredentialsAsXmlCommand.java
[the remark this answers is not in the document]
amuniz: Fixed. I found some more information here BTW https://github.com/jenkinsci/credentials-plugin/blob/master/docs/user.adoc

src/test/java/com/cloudbees/plugins/credentials/cli/CLICommandsTest.java
amuniz: Done.

amuniz:
@jtnord added the test and the folder credentials selector example.

About the format for the store in the CLI, I added a comment somewhere but it is collapsed now, so I copy here: "Totally agree. I had to deep dive in the code to guess what the `system::system::jenkins` string means (and I'm still not sure). This is the way all CLI commands in this plugin work. It is not my goal to change that now, but I can add an example for a folder, sure."

amuniz:
And BTW I found some more information about that weird format here: https://github.com/jenkinsci/credentials-plugin/blob/master/docs/user.adoc

amuniz:
@stephenc docs updated.

docs/user.adoc
```diff
@@ -1142,6 +1142,90 @@ c5e80c99-d7c3-4acb-a5e1-73592ceebbbc fred/******
...
+          <secret-redacted/>
+        </password>
+      </com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl>
+    </credentials>
+  </com.cloudbees.plugins.credentials.domains.DomainCredentials>
+</list>
+----
+
+The output of the command can be used as input (for --import) as is, the only needed change is to set the actual
+Secrets which are redacted in the output.
```
stephenc: How would a user have any idea that this “list” command can mutate?

How can a Jenkins admin lock down mutation while allowing read (eg if the Jenkins admin uses an init groovy script to filter out mutation CLI commands from the CLI command extension list.

I think import should be a separate command added to this PR

amuniz: > How would a user have any idea that this “list” command can mutate?

```
$> java -jar jenkins-cli.jar -noKeyAuth -s http://localhost:8080/jenkins/ help list-credentials-as-xml
java -jar jenkins-cli.jar list-credentials-as-xml STORE [--import]
Import and Export credentials as XML. The output of the command can be used as input (for --import) as is, the only needed change is to set the actual Secrets which are redacted in the output.
 STORE    : Store ID. For example the Jenkins credentials global store would be
            identified by "system::system::jenkins", and folder scoped
            credentials as "folder::item::/full/name/of/folder"
 --import : Read credentials as XML from the standard input and create them in
            the given store.
```

> How can a Jenkins admin lock down mutation while allowing read

Using permissions?

> if the Jenkins admin uses an init groovy script to filter out mutation CLI commands from the CLI command extension list.

Never saw such an exotic way to do that, but ok.

> I think import should be a separate command added to this PR

Ok, if you feel strongly... but I don't see the point.

amuniz:
@stephenc done.

amuniz:
@stephenc any plan on merging?

PR #102, opened 2018-06-26
[JENKINS-44772] check also for user scoped credentials + test

pskopek:
This PR is to fix https://issues.jenkins-ci.org/browse/JENKINS-44772
It also includes user scoped credential precedence over other scopes.

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
@@ -887,32 +887,33 @@ public abstract class CredentialsProvider extends Descriptor<CredentialsProvider
                     isParameter = true;
                     isDefaultValue = ((CredentialsParameterValue) parameter).isDefaultValue();
                     id = ((CredentialsParameterValue) parameter).getValue();
+                    if (id == null)
+                        return null;
                 }
             }
         }
+        List<C> jobContextCandidates = new ArrayList<>();
         // non parameters or default parameter values can only come from the job's context
         if (!isParameter || isDefaultValue) {
             // we use the default authentication of the job as those are the only ones that can be configured
             // if a different strategy is in play it doesn't make sense to consider the run-time authentication
             // as you would have no way to configure it
             Authentication runAuth = CredentialsProvider.getDefaultAuthenticationOf(run.getParent());
-            List<C> candidates = new ArrayList<C>();
             // we want the credentials available to the user the build is running as
-            candidates.addAll(
+            jobContextCandidates.addAll(
                     CredentialsProvider.lookupCredentials(type, run.getParent(), runAuth, domainRequirements)
             );
             // if that user can use the item's credentials, add those in too
             if (runAuth != ACL.SYSTEM && run.getACL().hasPermission(runAuth, CredentialsProvider.USE_ITEM)) {
-                candidates.addAll(
+                jobContextCandidates.addAll(
                         CredentialsProvider.lookupCredentials(type, run.getParent(), ACL.SYSTEM, domainRequirements)
                 );
             }
-            return CredentialsMatchers.firstOrNull(candidates, CredentialsMatchers.withId(id));
         }
+        List<C> candidates = new ArrayList<>();
         // this is a parameter and not the default value, we need to determine who triggered the build
         final Map.Entry<User, Run<?, ?>> triggeredBy = triggeredBy(run);
         final Authentication a = triggeredBy == null ? Jenkins.ANONYMOUS : triggeredBy.getKey().impersonate();
-        List<C> candidates = new ArrayList<C>();
         if (triggeredBy != null && run == triggeredBy.getValue()
                 && run.getACL().hasPermission(a, CredentialsProvider.USE_OWN)) {
             // the user triggered this job directly and they are allowed to supply their own credentials, so
@@ -937,6 +938,8 @@ public abstract class CredentialsProvider extends Descriptor<CredentialsProvider
                 );
             }
         }
+        // we need to add job's context candidates later to let user's credentials with the same id win
+        candidates.addAll(jobContextCandidates);
         C result = CredentialsMatchers.firstOrNull(candidates, CredentialsMatchers.withId(id));
         // if the run has not completed yet then we can safely assume that the credential is being used for this run
         // so we will track it's usage. We use isLogUpdated() as it could be used during post production
--- a/src/test/java/com/cloudbees/plugins/credentials/FindUserScopeCredentialsByIdTest.java
+++ b/src/test/java/com/cloudbees/plugins/credentials/FindUserScopeCredentialsByIdTest.java
@@ -1,0 +1,133 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc..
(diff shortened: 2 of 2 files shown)


anandka:
When will this be merged into master?

jcsou:
Hello,

I fork and use your update for this plugin.
Nominaly, a job can retrive user scoped credentials... Very usefull and very good for me.

My use case is running one job and, if it ok,  call another job.
I would like to propagate credential.

If credential is "scope global", job1 then job2 is ok ( credential use in job1 is propagetable by name in job2)
But, if credential is "scope user", job1 is ok, but job2 can't retrive credential to use...

Hope this help...

pskopek:
@jcsou do you have a test case for this scenario? Or are you able to create one?

jcsou:
Hi,

Considere two jenkinsfile : 
[TestSshJob1.jenkinsfile](https://github.com/jcsou/credentials-plugin/blob/user-scoped-creds/UserCredTest/TestSshJob1.jenkinsfile)
[TestSshJob2.jenkinsfile](https://github.com/jcsou/credentials-plugin/blob/user-scoped-creds/UserCredTest/TestSshJob2.jenkinsfile)

These jobs use credential "SSHUsernameWithPrivateKey" to connect to a server with ssh and sshagent.
Credential is a job parameter for them.

Adapte sshUserAtServer with a custom destination server or comment sh connection in sshagent section.

Mount these jenkins files as jobs in jenkins (TestSshJob1 and TestSshJob2)
Define SSHUsernameWithPrivateKey credentials : one scoped global and one scoped user.

Running TestSshJob2 alone with credentials globel or user is OK
Running TestSshJob1 with "global scoped credentials" is OK (TestSshJob2 run sucessfully)

Running TestSshJob1 with "user scoped credentials" :  TestSshJob2 **failed** ...
message is 
**FATAL: [ssh-agent] Could not find specified credentials**

I hope this understandable...

src/test/java/com/cloudbees/plugins/credentials/FindUserScopeCredentialsByIdTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    @Before
+    public void setUp() throws Exception {
+        final User alice = User.get("alice");
+        r.jenkins.setSecurityRealm(r.createDummySecurityRealm());
+        CredentialsStore userStore;
+        SecurityContext ctx = ACL.impersonate(alice.impersonate());
```
jglick: Use `ACL.as` in a `try`-with-resources block.

src/test/java/com/cloudbees/plugins/credentials/FindUserScopeCredentialsByIdTest.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Test
+    public void findUserCredentialsById() throws Exception {
+        final User alice = User.get("alice");
+        SecurityContext ctx = ACL.impersonate(alice.impersonate());
+        FreeStyleProject project = r.createFreeStyleProject();
+        project.getBuildersList().add(new PasswordBuildStep(testedCredentialId));
+        FreeStyleBuild build = project.scheduleBuild2(0, new Cause.UserIdCause(), new Action[0]).get();
+        logger.debug(JenkinsRule.getLog(build));
```
jglick: Just use `BuildWatcher`.

src/test/java/com/cloudbees/plugins/credentials/FindUserScopeCredentialsByIdTest.java
```diff
@@ -0,0 +1,133 @@
...
+        r.assertBuildStatus(Result.SUCCESS, build);
```
jglick:
```java
FreeStyleBuild build = r.buildAndAssertSuccess(project);
```

src/test/java/com/cloudbees/plugins/credentials/FindUserScopeCredentialsByIdTest.java
```diff
@@ -0,0 +1,133 @@
...
+                listener.getLogger().printf("Extracted secret: %s%n", password.getPlainText());
+            }
+            return true;
+        }
+
+        @TestExtension
+        public static class DescriptorImpl extends Descriptor<Builder> {
+
+            @Nonnull
+            @Override
```
jglick: Can just omit this override, assuming a sufficiently new `jenkins.version`.

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -887,32 +887,33 @@ public void remove() {
...
                     CredentialsProvider.lookupCredentials(type, run.getParent(), runAuth, domainRequirements)
             );
             // if that user can use the item's credentials, add those in too
             if (runAuth != ACL.SYSTEM && run.getACL().hasPermission(runAuth, CredentialsProvider.USE_ITEM)) {
-                candidates.addAll(
+                jobContextCandidates.addAll(
                         CredentialsProvider.lookupCredentials(type, run.getParent(), ACL.SYSTEM, domainRequirements)
                 );
             }
-            return CredentialsMatchers.firstOrNull(candidates, CredentialsMatchers.withId(id));
```
stephenc: I'm getting scared by this.

The intent (as documented by the comment on line 893) is that non-parameters and default parameter values can only come from the job's context and not from the user's context.

Now it may be that we need to revisit that... but there are implications in opening it up.

wfxr:
Any progress?

jglick:
The desired design here is not at all clear, and this PR does not even include a functional test with Pipeline (the use case in JENKINS-44772). There have been recent discussions about the broader goal, and I am not convinced the workflow described in JIRA is even something we want. Improper approaches could also be security risks, and even the creator of Credentials is unsure whether this patch is right. Therefore I do not think this PR should proceed, at least without much more serious analysis.

matthiasbaldi:
@jglick I can see your topic.
But, what is now happening with this issue?
This was a working feature before an Jenkins upgrade and we used it.
So, my question is, when would we have a working user credential feature back?
Or can we solve this problem with an other plugin?

jvz:
I have an alternative proposal in #119 which avoids hard-coding user-scoped credential ids. After playing around with that feature, I wouldn't even recommend giving your user-scoped credentials human readable ids and instead stick with the randomly generated ones.

EronWright:
Reporting that the following workaround was effective for me: [JENKINS-38963/comment](https://issues.jenkins-ci.org/browse/JENKINS-38963?focusedCommentId=318189&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-318189)

The `credentialsId` parameter is very finicky; it must be of the form `${PARAMETER-NAME}`.
- Good: `credentialsId: '${MYCREDENTIALS}'`
- Bad: `credentialsId: '${params.MYCREDENTIALS}'`
- Bad: `credentialsId: params.MYCREDENTIALS`

jvz:
Closing this in favor of the approach developed in #119 and the related ticket, [JENKINS-58170](https://issues.jenkins-ci.org/browse/JENKINS-58170).

PR #106, opened 2018-10-08
Add i18n support for credentials list view

LinuxSuRen:


LinuxSuRen:
@jvz Do anything hold on this PR?

jvz:
Same here @dwnusbaum @jglick

src/main/resources/com/cloudbees/plugins/credentials/ViewCredentialsAction/index.jelly
```diff
@@ -85,7 +85,7 @@
                    class='model-link inside'>
                   <j:choose>
                     <j:when test="${d.domain.global}">
-                      (global)
+                      ${%(global)}
```
jglick: This syntax does not work. Nothing appears, since `(`…`)` is the syntax to introduce a parameterized bundle key.

You would need to create an `index.properties` like

```properties
global=(global)
```

and use Jelly

```
${%global}
```

Anyway, try it.

LinuxSuRen: You're right.

LinuxSuRen:
No ideas why this build is not success.

jglick:
By the way please avoid force pushes inside PRs since it breaks the GitHub feature of seeing “changes since your last review” which I find valuable. A committer can always decide to squash-merge if the commit history is a mess and they care.

jglick:
Trying to trigger a fresh build. There was a CI problem (the Windows nodes are flaky).

LinuxSuRen:
@jglick Got it. I did that just for a better commit log. But your suggestion is fair enough. I'll take it. And thanks for that.

PR #107, opened 2018-11-27
Update UserCredentialsProvider.java

liucc52:
User.current() will always return SYSTEM during running pipeline steps, this will cause checkpermission failing for pipeline step because the login user will not be SYSTEM


Wadeck:
@liucc52 If you are still encountering this bug, could you please file a Jira ticket explaining a bit more in details the problem? From my PoV it was working since several years and haven't seen the behavior you described.

PR #108, opened 2019-02-04
Sentences should end with periods...

jsoref:


jvz:
Third time's the charm!

PR #109, opened 2019-02-05
replace j:local with j:scope

jsoref:
Per https://github.com/jenkinsci/credentials-plugin/pull/62#discussion_r254006084


jsoref:
@stephenc

PR #112, opened 2019-04-10
Java 11 readiness: build both on JDK8 and JDK11

batmat:
Make sure this plugin is constantly checked both building with JDK8 and JDK11.

This goes with the Jenkins Java 11 General Availability announcement made back in March, and we're making a pass to check plugins are looking good on a JDK11.

([to understand the Jenkinsfile content](https://wiki.jenkins.io/display/JENKINS/Java+11+Developer+Guidelines#Java11DeveloperGuidelines-MakesureyourpluginistestedinContinuousIntegrationonJava8andJava11atthesametime))


batmat:
```
[INFO] --- access-modifier-checker:1.16:enforce (default-enforce) @ credentials ---
[ERROR] com/cloudbees/plugins/credentials/cli/BaseCredentialsCLICommand:100 jenkins/util/xml/XMLUtils must not be used
[ERROR] com/cloudbees/plugins/credentials/CredentialsStoreAction$CredentialsWrapper:1514 jenkins/util/xml/XMLUtils must not be used
[ERROR] com/cloudbees/plugins/credentials/CredentialsStoreAction$DomainWrapper:815 jenkins/util/xml/XMLUtils must not be used
[ERROR] com/cloudbees/plugins/credentials/CredentialsStoreAction$DomainWrapper:1012 jenkins/util/xml/XMLUtils must not be used
[ERROR] com/cloudbees/plugins/credentials/CredentialsStoreAction:549 jenkins/util/xml/XMLUtils must not be used
```

MRamonLeon:
There is an issue asking for the removal of this dependency: https://issues.jenkins-ci.org/browse/JENKINS-52600

MRamonLeon:
I've copied the restricted classes and avoid a dereference warning of spotbugs.
@batmat @alecharp @Wadeck @daniel-beck @fcojfernandez @oleg-nenashev  @reviewbybees

pom.xml
```diff
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.1</version>
+    <version>3.42</version>
```
alecharp: could be `3.43` now. If I recall correct, the next version has a `javadoc` fix as well, no?

MRamonLeon:
I think we need a close-reopen

batmat:
LGTM

batmat:
I think it's ready for review by maintainers. Just requested review from the last persons I see merged PRs in last commits.

MRamonLeon:
There is a new PR on core to unrestrict XMLUtils (https://github.com/jenkinsci/jenkins/pull/4032). When that is released, we can remove the new util package and restore the dependency with the core XMLUtils.

MRamonLeon:
I've filed the task to do that: https://issues.jenkins-ci.org/browse/JENKINS-57518 later

pom.xml
```diff
@@ -65,8 +65,8 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.625.3</jenkins.version>
-    <java.level>7</java.level>
+    <jenkins.version>2.130</jenkins.version>
```
jglick: Why not just the next LTS?
```suggestion
    <jenkins.version>2.138.4</jenkins.version>
```

pom.xml
```diff
@@ -140,13 +140,6 @@
...
-      <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>findbugs-maven-plugin</artifactId>
-        <configuration>
-          <excludeFilterFile>src/findbugs/excludesFilter.xml</excludeFilterFile>
-        </configuration>
-      </plugin>
       <plugin>
         <groupId>org.antlr</groupId>
         <artifactId>antlr4-maven-plugin</artifactId>
```
jglick: Apparently there is a “Credentials Query Language” from #55 (no code review or downstream PRs with usage examples). @stephenc?

----

```java
(instanceof com.cloudbees.plugins.credentials.common.UsernameCredentials) && !(username == "bob")
```

----

This walk-through is incomplete and needs the following additions:

* [ ] …
* [ ] Provide some details on how the Credentials Query Language can be used to limit querying credentials from the remote service

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -887,6 +887,8 @@ public void remove() {
                     isParameter = true;
                     isDefaultValue = ((CredentialsParameterValue) parameter).isDefaultValue();
                     id = ((CredentialsParameterValue) parameter).getValue();
+                    // Avoid spotbugs complaining about the id being null and used lately on CredentialsMatchers.withId
```
jglick: Is this fixing a real bug, or do we just want to `@SuppressFBWarnings`?

src/main/java/com/cloudbees/plugins/credentials/common/AbstractIdCredentialsListBoxModel.java
```diff
@@ -67,15 +67,15 @@
  * public String getCredentialsId() {return credentialsId;}
  * </pre>
  * <p>
- * Your <tt>config.groovy</tt> should have the following entry to render a drop-down list box:
+ * Your <code>config.groovy</code> should have the following entry to render a drop-down list box:
```
jglick:
```suggestion
 * Your {@code config.groovy} should have the following entry to render a drop-down list box:
```

src/main/java/com/cloudbees/plugins/credentials/common/AbstractIdCredentialsListBoxModel.java
```diff
@@ -67,15 +67,15 @@
...
  * </p>
  * <pre>
  * f.entry(title:_("Credentials"), field:"credentialsId") {
  *     f.select()
  * }
  * </pre>
  * <p>
- * Finally, your {@link Descriptor} implementation should have the <tt>doFillCredentialsIdItems</tt> method, which
+ * Finally, your {@link Descriptor} implementation should have the <code>doFillCredentialsIdItems</code> method, which
```
jglick:
```suggestion
 * Finally, your {@link Descriptor} implementation should have the {@code doFillCredentialsIdItems} method, which
```

src/main/java/com/cloudbees/plugins/credentials/common/AbstractIdCredentialsListBoxModel.java
```diff
@@ -533,7 +533,7 @@
      * list is nonempty.)
      *
      * @param c collection containing elements to be added to this list
-     * @return <tt>true</tt> if this list changed as a result of the call
+     * @return <code>true</code> if this list changed as a result of the call
```
jglick:
```suggestion
     * @return {@code true} if this list changed as a result of the call
```

src/main/java/com/cloudbees/plugins/credentials/util/RestrictiveEntityResolver.java
```diff
@@ -0,0 +1,32 @@
...
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.xml.sax.EntityResolver;
+import org.xml.sax.InputSource;
+import org.xml.sax.SAXException;
+
+import java.io.IOException;
+
+/**
+ * This class is a clone of {@link jenkins.util.xml.RestrictiveEntityResolver} because the last is Restricted. If we
+ * ever unrestrict it, we should remove this class and use the first one.
```
jglick: So, file a core PR to unrestrict it, since clearly plugins need it.

pom.xml
batmat: Because it was the minimum required to make the tests succeed, but I agree.

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
MRamonLeon: AFAICT `getValue` can return a `null` and the constructor of a `CredentialsParameterValue` allows you to pass a `null`. Maybe the current use doesn't imply a bug, but a future use in tests or whatever could land on a real bug.

src/main/java/com/cloudbees/plugins/credentials/util/RestrictiveEntityResolver.java
MRamonLeon: We already did

jglick: https://github.com/jenkinsci/jenkins/pull/4032, I see. So amend this comment to include the magic `TODO` keyword and specify the (weekly) version containing that API (https://github.com/jenkinsci/jenkins/pull/4032#discussion_r287387863). That way future maintainers can just grep for TODOs and quickly see if there is something actionable.

MRamonLeon: done

oleg-nenashev:
@stephenc @dwnusbaum Would you be fine if we merge this change?
CC @jenkinsci/java11-support

pom.xml
stephenc: Yes there is a query language designed to allow remote credentials stores to optimise search scope to limit trawling through all the credentials.

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
stephenc: So the initial credentials api did not mandate an ID because the idea was to use it as a marker interface. Very quickly we discovered that wasn’t great because everyone had to implement their own select UI... so then we added IdCredentials to enforce that.

In theory, someone could have a plugin depending on the 2012-era api and not having an ID... those would be `null` but of limited utility as you cannot integrate them widely.

At some point we should declare a major version bump and move ID to the main interface and deprecating IdCredentials. I do not think today is that day 😂

batmat:
@jglick @stephenc given you approved and many others did, that would be great if you can consider merging.
Thanks!

batmat:
Thanks Stephen

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
jvz: Oh I'd love to be onboard with a version 2 plugin sometime. We can remove all the deprecated stuff, merge interface extensions back together, and perhaps make other API improvements.

PR #113, opened 2019-04-18
Avoid confusion between UUID and ID

Wadeck:
This information is used multiple times, automatically in the jenkins.io doc.

@daniel-beck


daniel-beck:
Autogenerated IDs are UUIDs, and really old versions of the plugin did not allow user-specified IDs, so this is probably a leftover from that.

PR #114, opened 2019-04-24
[JENKINS-57559] - Add credentials configurator for configuration as code

casz:
Thanks @batmat for starting the work to get this ready for JDK 8 and JDK 11.

I really want to get credentials configurator out of JCasC support!

Regarding XMLUtils usage and XML request parsing.
Should we remove the XML code?
This search shows that we are the only plugin that parses XML request
[org:jenkinsci "getContentType" xml extension:java path:src/main](https://github.com/search?p=2&q=org%3Ajenkinsci+%22getContentType%22+extension%3Ajava+xml+path%3Asrc%2Fmain&type=Code)


pom.xml
```diff
@@ -65,9 +65,11 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.625.3</jenkins.version>
-    <java.level>7</java.level>
+    <jenkins.version>2.60.3</jenkins.version>
+    <java.level>8</java.level>
     <antlr4.version>4.5</antlr4.version>
+    <jcasc.version>1.13</jcasc.version>
+    <access-modifier-checker.failOnError>false</access-modifier-checker.failOnError>
```
casz: I would like to avoid this :sweat: Can we remove the XML parsing?

casz:
:cry: Jenkins doesn't trust me to use JDK 11 build but at least it passes on JDK 8!

casz:
https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fcredentials-plugin/detail/PR-114/2/pipeline/8

casz:
@stephenc I know your busy but could you weigh in on XML removal?

src/test/java/com/cloudbees/plugins/credentials/casc/SystemCredentialsTest.java
```diff
@@ -0,0 +1,98 @@
...
+        Mapping domain = domainCredentials.get("domain").asMapping();
+        Sequence credentials = domainCredentials.get("credentials").asSequence();
+        Mapping usernamePassword = credentials.get(0).asMapping().get("usernamePassword")
+            .asMapping();
+        Mapping hostnameSpecification = domain.get("specifications").asSequence().get(0).asMapping()
+            .get("hostnameSpecification").asMapping();
+
+        assertThat(usernamePassword.getScalarValue("scope"), is("SYSTEM"));
+        assertThat(usernamePassword.getScalarValue("username"), is("root"));
+        assertThat(usernamePassword.getScalarValue("password"), is(not("password")));
```
timja: you could also decrypt it with Secret.decrypt to check what is there is an encrypted string?

casz:
I should do a downstream PR on JCasC

casz:
@jenkinsci/code-reviewers @stephenc @batmat anyone cares to weigh in?

pom.xml
daniel-beck: > Can we remove the XML parsing?

Context?

casz: I didn't look too closely. It seems that credentials allow you to send both JSON and XML requests to create credentials.

daniel-beck: Which API are you using that's prohibited to plugins?

casz: XMLUtils is marked restricted and is used here:
https://github.com/jenkinsci/credentials-plugin/blob/6f710a7e30156e94aff8a3c6c6799cef04ad35df/src/main/java/com/cloudbees/plugins/credentials/CredentialsStoreAction.java
https://github.com/jenkinsci/credentials-plugin/blob/6f710a7e30156e94aff8a3c6c6799cef04ad35df/src/main/java/com/cloudbees/plugins/credentials/cli/BaseCredentialsCLICommand.java

daniel-beck: Ah, OK, it used to be available due to a bug in the access modifier checker used in old parent POMs. We're not updating that, so can no longer use it by default. That explains why it didn't show in the diff.

casz: So your saying we should keep the code as is?
That would mean that this code is ready for merging I'd guess.

src/main/java/com/cloudbees/plugins/credentials/casc/CredentialsRootConfigurator.java
```diff
@@ -0,0 +1,94 @@
...
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+import java.util.Collections;
+import java.util.Set;
+
+import static io.jenkins.plugins.casc.Attribute.noop;
+
+
+/**
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
```
stephenc: No author tags, and real javadoc comments for all non override methods (protected and above) and all fields (private and above)

casz: I'll fix those up

stephenc:
> Regarding XMLUtils usage and XML request parsing.
> Should we remove the XML code?

IIRC the XML request stuff is documented in the User Guide. As such to remove, you would need a major version bump and an update to the user guide. I suspect that is well outside the scope of this change, better would be to:

1. Copy the required bits of `XMLUtils` in tree, mark as restricted. Done with this PR

2. In a new PR, every use of the copy of `XMLUtils` should capture the request source in a size-bounded hashmap with the source being the key and the value being the number of requests.
3. An administrative monitor should alert if the map is non-empty and provide the details of the source and the number of requests and that a subsequent version of the plugin may disable by default.
4. A system property controls whether to even let the in-tree `XMLUtils` request go through or blow up with a run-time exception.

It's not so much that we want to stop using the XML request type, rather than we don't want to have to maintain our own fork in-tree... but we have to give users a migration path. Then we follow up with a major version bump that switches the system property on by default.

pom.xml
stephenc: :-1: we should just copy the bits we need in-tree. Turning off this checker is not the solution

casz: done :+1:

src/main/java/com/cloudbees/plugins/credentials/casc/CredentialsRootConfigurator.java
casz: I went through and removed author tags as asked. You had a couple yourself :relaxed:

casz: Done, done :+1:

src/test/java/com/cloudbees/plugins/credentials/casc/SystemCredentialsTest.java
casz: Done :+1:

casz:
@stephenc I have addressed your concern :blush: Hopefully we can move forward!

stephenc:
https://github.com/jenkinsci/credentials-plugin/blob/master/docs/user.adoc Should at least include an example of how to use this new feature. Otherwise :100:

stephenc:
To be clear, the docs can be anemic, but there must be docs.

stephenc:
In other words, to all i an asking for is:

1. A heading
2. Two or three sentences 
3. A code snippet block with an example config as code

If you want to go into more details, that’s fine... but could also be left for a follow up PR.

The idea is: if we have a feature documented, people at least know where to add improvements... if we don’t document it, it’s not a feature. I put a significant amount of effort into ensuring all the functionality is at least mentioned somewhere in the docs, so I don’t want gaps evolving as it takes a lot of work to find gaps after they have evolved

casz:
Thanks for the ample reasoning 😆 I will make it so, I'll try my best at least.  
Hopefully, some kind-hearted reviewer will look at it 😉

oleg-nenashev:
Created https://issues.jenkins-ci.org/browse/JENKINS-57559 to track this change

amuniz:
This is looking good to me! :+1:

src/main/java/com/cloudbees/plugins/credentials/xml/XMLUtils.java
```diff
@@ -0,0 +1,101 @@
...
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.xml.sax.InputSource;
+import org.xml.sax.SAXException;
+import org.xml.sax.XMLReader;
+import org.xml.sax.helpers.XMLReaderFactory;
+
+/**
+ * Utilities useful when working with various XML types.
+ */
+@Restricted(NoExternalUse.class)
```
jtnord: Why?  this is a brutal yank from core and is missing the copyright header.
we can just use the one from core (we already had - just ignore the restriction temporarily as it is removed in the latest master)
At the very least this should include a TODO to remove when the baseline is > whatever Fridays release will be.

casz:
Thanks to @timja for doing the doc :sweat_smile:

pom.xml
```diff
@@ -119,6 +128,19 @@
       <version>2.2.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>cloudbees-folder</artifactId>
```
casz: Fixed the misaligned indent
https://github.com/jenkinsci/credentials-plugin/blob/37841b8970cfc856d6463739b86fffe4792944dd/pom.xml#L116-L127

casz:
@oleg-nenashev @stephenc @daniel-beck hopefully this should be easier to review now :sweat_smile: 

Windows agent strikes again @olblak @rtyler :cry: 
https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fcredentials-plugin/detail/PR-114/15/pipeline#log-147

casz:
Nope, we are all good :raised_hands:

casz:
@stephenc what are the chances for a merge? We have a working role-strategy plugin and credentials plugin.

Meaning we could completely get rid of the support plugin anytime now 🎉
In other words, once role-strategy and credentials plugin has released, we can move forward in JCasC plugin and get rid of the support plugin.

casz:
🙌 THANK YOU!

casz:
@dwnusbaum @jglick @stephenc would appreciate a release 😅 Currently "somewhat" blocking a release of JCasC cause I want to include the removal of support plugin in the next release.

oleg-nenashev:
I can cut the release if @stephenc blesses it

PR #117, opened 2019-05-29
update antlr from 4.5 to 4.5.3

based2:


jeffret-b:
Hmmm ... I'd kind of like to see us just move forward to the latest instead of just a tiny step forward. But, the latest breaks tests. (Currently 4.7.2.) Version 4.6 works fine though. Tests start breaking with 4.7. Not sure if it would be best to move to 4.7.2 and fix issues or just accept 4.6.

jvz:
I'm working on updating this to 4.9.1 now.

PR #119, opened 2019-06-12
[JENKINS-58170] Allow credential parameters to shadow credential ids in lookup

jvz:
See [JENKINS-58170](https://issues.jenkins-ci.org/browse/JENKINS-58170) for a longer description.

This PR addresses both that issue as well as an API to support changes in pipeline-input-step-plugin to fix [JENKINS-47699](https://issues.jenkins-ci.org/browse/JENKINS-47699).

Downstream: https://github.com/jenkinsci/pipeline-input-step-plugin/pull/36.

Review requested: @daniel-beck @Wadeck @jeffret-b @jglick @stephenc @cyrille-leclerc

### DIFF
--- a/src/main/resources/lib/credentials/select.jelly
+++ b/src/main/resources/lib/credentials/select.jelly
@@ -75,6 +75,17 @@
     </j:if>
     <div field-disabled="${paramValue?'true':null}"
          class="credentials-select-content-select ${paramValue?'credentials-select-content-inactive':'credentials-select-content-active'}">
+      <!-- TODO: help links -->
+      <!-- TODO: update the build-with-parameters.png screenshot in documentation -->
+      <div class="include-user-credentials">
+        <div class="warning user-credentials-caution" style="display:none">
+          ${%userCredentialsCaution}
+        </div>
+        <label field-disabled="true">
+          <input type="checkbox" name="includeUser"/>
+          ${%includeUserCredentials}
+        </label>
+      </div>
       <f:select clazz="${attrs.clazz} credentials-select" field="${attrs.field}"
                 default="${paramDefault?'':attrs.default}" value="${paramValue?(paramDefault?'':attrs.default):value}"
                 checkMethod="${attrs.checkMethod}"/>
--- a/src/main/resources/lib/credentials/select.properties
+++ b/src/main/resources/lib/credentials/select.properties
@@ -25,3 +25,5 @@ invalidExpression=A valid parameter expression consists of the parameter name en
   within <code>$'{'</code> and <code>'}'</code>
 fixedSelection=Specific credentials
 paramExpression=Parameter expression
+includeUserCredentials=List user credentials
+userCredentialsCaution=Only provide user credentials to trusted builds
--- a/src/main/resources/lib/credentials/select/select.js
+++ b/src/main/resources/lib/credentials/select/select.js
@@ -260,6 +260,18 @@ Behaviour.specify("INPUT.credentials-select", 'credentials-select', -100, functi
   }).bind($(x));
   x.onchange();
 });
+Behaviour.specify("DIV.include-user-credentials", 'include-user-credentials', 0, function (e) {
+    var includeUserCheckbox = e.querySelector("input[name='includeUser']");
+    var caution = e.querySelector("div.user-credentials-caution");
+    includeUserCheckbox.addEventListener('click', function (evt) {
+        if (this.checked) {
+            caution.setAttribute('style', 'display:block');
+        } else {
+            caution.setAttribute('style', 'display:none');
+        }
+        layoutUpdateCallback.call();
+    });
+})
 window.setTimeout(function() {
     // HACK: can be removed once base version of Jenkins has fix of https://issues.jenkins-ci.org/browse/JENKINS-26578
     // need to apply the new behaviours to existing objects
--- a/src/test/java/com/cloudbees/plugins/credentials/CredentialsParametersActionTest.java
+++ b/src/test/java/com/cloudbees/plugins/credentials/CredentialsParametersActionTest.java
@@ -1,0 +1,50 @@
+package com.cloudbees.plugins.credentials;
+
+import com.cloudbees.plugins.credentials.common.IdCredentials;
+import com.cloudbees.plugins.credentials.domains.Domain;
+import com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl;
+import hudson.model.Cause;
+import hudson.model.FreeStyleBuild;
+import hudson.model.FreeStyleProject;
+import hudson.model.ParametersAction;
+import hudson.model.ParametersDefinitionProperty;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import java.io.IOException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+public class CredentialsParametersActionTest {
+
+    @ClassRule public static JenkinsRule j = new JenkinsRule();
+
+    @BeforeClass
+    public static void setUpClass() throws IOException {
+        j.jenkins.setSecurityRealm(j.createDummySecurityRealm());
+        CredentialsProvider.lookupStores(j.jenkins).iterator().next()
+                .addCredentials(Domain.global(), new UsernamePasswordCredentialsImpl(CredentialsScope.GLOBAL, "cred-id", "global credential", "root", "correct horse battery staple"));
+    }
+
+    @Test
+    public void forRunReturnsNullWhenRunHasNoParameters() throws Exception {
+        final FreeStyleProject project = j.createFreeStyleProject();
+        final FreeStyleBuild build = j.assertBuildStatusSuccess(project.scheduleBuild2(0));
+        assertNull(CredentialsParametersAction.forRun(build));
+    }
+
+    @Test
+    public void forRunCopiesParametersWhenRunIsParameterized() throws Exception {
+        final FreeStyleProject project = j.createFreeStyleProject();
+        project.addProperty(new ParametersDefinitionProperty(new CredentialsParameterDefinition("cred", "", null, IdCredentials.class.getName(), true)));
+        final CredentialsParameterValue parameterValue = new CredentialsParameterValue("cred", "cred-id", "", false, "alpha");
+        final FreeStyleBuild build = j.assertBuildStatusSuccess(project.scheduleBuild2(0, new Cause.UserIdCause("alpha"), new ParametersAction(parameterValue)));
+        final CredentialsParametersAction action = CredentialsParametersAction.forRun(build);
+        assertNotNull(action);
+        assertEquals(parameterValue, action.findParameterByName("cred"));
+    }
+}
--- a/src/test/java/com/cloudbees/plugins/credentials/CredentialsParametersActionTest.java
+++ b/src/test/java/com/cloudbees/plugins/credentials/CredentialsParametersActionTest.java
@@ -1,3 +1,27 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2019 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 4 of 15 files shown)


vilacides:
@daniel-beck @Wadeck @jeffret-b reviews are appreciated!

jvz:
@jenkinsci/java11-support is there a workaround for this Java 11 javadoc error? It's making this project impossible to verify through CI right now.

batmat:
@jvz I would simply recommend getting rid of this construct.

There might be a debate around the fact if this should be allowed or not, because technically `@link` allows a class, and a `byte[]` _is_ a class as per the [spec](https://docs.oracle.com/javase/specs/jls/se7/html/jls-10.html#jls-10.8).

But I do not think this is worth our time trying to get around it.

What puzzles me slightly more is how the recent-ish builds on master were successful (end of May). And I do not think the JDK got upgraded after this.

Anyway, I would recommend fixing this in a separate PR, get a green build there, approve, merge, and rebuild here. This so we separate concerns and accelerate the process overall. (This is unrelated to the change at stake **here**).

jvz:
Thanks, @batmat, for the quick fix!

jvz:
I agree that it looks cramped. I didn't add any styling. Maybe there's some UI guidelines somewhere?

Wadeck:
> Maybe there's some UI guidelines somewhere?

You do what you can, there is no guideline that I am aware of

Wadeck:
> way things shift after clicking

In order to keep the mental flow from top to bottom, putting the warning after the checkbox seems more natural.

<details>

<summary>Gifs before/after</summary>

Current version:
user_scope_creds_warning_current

Proposal:
user_scope_creds_warning_proposal




</details>

jvz:
@Wadeck that's a much better idea, and I was also amusing myself with the shifting UI that was super annoying.

jeffret-b:
I'm not a UI guy, but it's better below -- I still find the shifting UI annoying.

Wadeck:
> the shifting UI annoying.

Another possibility is to "reserve" the space by using `visibility:hidden` and then `visibility:visible` instead of the `display:none` and `display:block`. Con: that will let a blank area.

Or putting the warning message in floating mode, with `float:right` and letting appearing/disappearing on the right, without touching the "flow". Con: on wide screen the space between checkbox and warning could be huge.

Or last proposal (my favorite), just putting the warning "inline" after the checkbox, continuing after some px of padding. Pro: no con? ^^

jeffret-b:
> Or last proposal (my favorite), just putting the warning "inline" after the checkbox, continuing after some px of padding. Pro: no con? ^^

This one makes the most sense to me (without having seen how it looks).

jvz:
I didn't try out the `visibility:hidden` yet, though adding some padding and putting the checkbox below made it visibly better so far.

jvz:
@Wadeck those are interesting ideas. I'll try them out, pick which looks right, and then update this PR.

src/main/resources/lib/credentials/select/select.js
```diff
@@ -260,6 +260,17 @@ Behaviour.specify("INPUT.credentials-select", 'credentials-select', -100, functi
...
 });
+Behaviour.specify("DIV.include-user-credentials", 'include-user-credentials', 0, function (e) {
+    e.querySelector("input[name='includeUser']").addEventListener('click', function (evt) {
+        var caution = e.querySelector("span.user-credentials-caution");
+        if (this.checked) {
+            caution.setAttribute('style', 'display:inline');
+        } else {
+            caution.setAttribute('style', 'display:none');
+        }
+        layoutUpdateCallback.call();
```
jvz: Now that the caution message is displayed inline, this might not be needed.

jvz:
* I ended up switching first to `inline-block`, then to `inline`, and finally switching it to a simple `span` instead of `div` which works out fine with a padding improvement.
* I've made that a feature flag attribute for `<c:select/>` instead of a global UI change.
* The `userId` field is now only used transiently.

Wadeck:
### Warning UI

Latest version with span from Matt:

<details>

<summary>Gif</summary>

user_scope_creds_warning_better_ui


</details>

Seems a lot better :)

---

### Credential theft vulnerability

The userId was removed from parameter in the stored file.

- [ ] Bob cannot replay the build from Alice if there was some user-scope credentials (and thus can steal them by modying the pipeline)
  ⚠️ **still vulnerable even after the transcient modification**


---

### Replay (non-regression)
- [x] Bob can still replay the build from Alice if there was just a global credentials
- [x] Alice can replay her previous build with/without user-scope credentials

---

### Job Configuration
- [x] The checkbox does not appear (neither the warning)
  (the user-scope credentials were not usable there but the checkbox was present without any effect)


### Delayed execution
Using Schedule Build Plugin
- [x] The credentials are displayed without information leakage
  Hopefully the toString method is not overriden in `CredentialsParameterValue`...

<details>

<summary>Screenshot</summary>

Screenshot_2019-07-24_104335_001


</details>
- [x] No way found to modify the build during the "waiting period" and thus to steal the credentials

daniel-beck:
> Using Schedule Build Plugin

There's a query parameter when you click "Build With Parameters", change that from `0sec` to `1000sec`.

jeffret-b:
I like the new UI layout better.

jvz:
> ### Credential theft vulnerability
> 
> The userId was removed from parameter in the stored file.
> 
>     * [ ]  Bob cannot replay the build from Alice if there was some user-scope credentials (and thus can steal them by modying the pipeline)
>       ⚠️ **still vulnerable even after the transcient modification**

I see this is because `ReplayAction` copies over the `ParametersAction`, so I'll move the user id data into this new action instead.

src/main/resources/lib/credentials/select/select.js
Wadeck: Please remove it if it's not needed :)

src/test/java/com/cloudbees/plugins/credentials/CredentialsParametersActionTest.java
```diff
@@ -0,0 +1,120 @@
...
+        assertEquals("cred-id", creds.getParameterValue().getValue());
+    }
+
+    @Test
+    public void forRunUsesUserIdCauseForInitialParametersOwnership() throws Exception {
+        final FreeStyleProject project = j.createFreeStyleProject();
+        project.addProperty(new ParametersDefinitionProperty(new CredentialsParameterDefinition("cred", "", null, IdCredentials.class.getName(), true)));
+
+        // no cause means no user id
+        {
```
Wadeck: @jeffret-b are you ok with this construct? I remember to have some comments from you about using methods instead of blocks in test.

src/main/resources/lib/credentials/select/select.js
```diff
@@ -260,6 +260,21 @@ Behaviour.specify("INPUT.credentials-select", 'credentials-select', -100, functi
...
+        var caution = e.querySelector('span.user-credentials-caution');
+        if (this.checked) {
+            caution.classList.remove('hidden');
+        } else {
+            caution.classList.add('hidden');
+        }
+    });
+    // simpler version of f:helpLink using inline help text
+    e.querySelector('a.help-btn').addEventListener('click', function (evt) {
+        this.remove();
```
jvz: Before anyone says "hey wait a minute, isn't this function a little too new to use across all browsers?", let me point out that having the question mark not remove itself after clicking it does no harm.

docs/user.adoc
```diff
@@ -1521,7 +1521,7 @@ The list of available credentials will depend on a number of factors:
 
 [WARNING]
 ====
-Credentials parameters can only access the per-user credentials store of the job that was explicitly triggered by the user.
+Credentials parameters can only access user credentials if the corresponding user is the one who chooses them during a build.
```
jeffret-b: I'm not sure what this is saying. The phrasing is a little convoluted. It could be unwound for clarity.

src/main/resources/lib/credentials/select.jelly
```diff
@@ -75,6 +80,27 @@
     </j:if>
     <div field-disabled="${paramValue?'true':null}"
          class="credentials-select-content-select ${paramValue?'credentials-select-content-inactive':'credentials-select-content-active'}">
+      <j:choose>
+        <j:when test="${includeUser and attrs.includeUserCaution}">
+          <!-- TODO: update the build-with-parameters.png screenshot in documentation -->
```
jeffret-b: Does this TODO need to be completed?

src/main/resources/lib/credentials/select.properties
```diff
@@ -24,4 +24,9 @@
 invalidExpression=A valid parameter expression consists of the parameter name enclosed \
   within <code>$'{'</code> and <code>'}'</code>
 fixedSelection=Specific credentials
-paramExpression=Parameter expression
\ No newline at end of file
+paramExpression=Parameter expression
+includeUserCredentials=List user credentials
+includeUserCredentialsHelp=Whether or not to include user credentials in the available \
```
jeffret-b:
```suggestion
includeUserCredentialsHelp=Whether to include user credentials in the available \
```
Very minor suggestion, but the "or not" is duplicative and takes up screen space.

src/main/resources/lib/credentials/select.jelly
jvz: Ideally, yes. I need to set up a test project to take a similar screenshot to the current one.

src/main/resources/lib/credentials/select/select.js
Wadeck: Thank you for the highlight here :)

Actually the `remove` function is not browser dependent, it's provided by prototype. The browser methods are `parentNode` and `removeNode` which are supported everywhere (even on IE6 🎉).
Side note, if there was a problem with that method, it will be harmful because the second line will not be executed and thus, the help block not being displayed.

---

:warning: But there is another problem here. I tested with IE and even with Chrome, the method seems to be called twice and thus, log a javascript error in the console. You cannot remove an element that was already removed.
```
Uncaught TypeError: Cannot read property 'removeChild' of null
    at remove (prototype.js:1992)
    at HTMLAnchorElement._methodized [as remove] (prototype.js:461)
    at HTMLAnchorElement.<anonymous> (select.js:274)
```

This problem is due to the hack at the bottom of the script (👿).

Suggestion: 
```suggestion
        this.parentNode && this.remove();
```

That should avoid the error to be displayed every time you click on the icon.

---

The "new" UI/UX introduced here is not consistent with the rest of the project.
1) The icon normally does not go away after a click
2) The icon should allow to re-click to close
3) Normally the icon is at the right of the screen
4) The icon is glued with the warning message as displayed in the following screenshot
5) There is no link to the plugin providing that info text (very useful when trying to debug a problem)

<details>

<summary>Screenshot</summary>

Screenshot_2019-08-23_132728_001


</details>

As `c:select` is not using the "standard" table layout used by the rest of the component, we cannot use the "standard" help here, thus we have troubles. With this change, we are going even further in the non-standard direction.

Personally I will not block the PR due to the UI inconsistency, but it's sad to have to use such hacky way just to have a help message. cc @josephbrueggen.

src/main/java/com/cloudbees/plugins/credentials/builds/CredentialsParameterBinder.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Tracks credentials being bound and unbound to a build. An instance is created and attached to a build when it is
+ * first looked up via {@link #getOrCreate(Run)}. This binds any existing {@link CredentialsParameterValue}s using the
+ * {@link hudson.model.Cause.UserIdCause} if available. Other plugins may
+ * {@linkplain #bindCredentialsParameter(String, CredentialsParameterValue) bind} and
+ * {@linkplain #unbindCredentialsParameter(String) unbind} parameters during a build.
+ *
+ * @since 2.3.0
```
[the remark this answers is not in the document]
jvz: I'm the only one making releases currently, so this was intentional.

jvz:
Got an incoming set of commits addressing nearly everything you mentioned, @Wadeck, besides the additional tests.

jvz:
Ok, it took a bit of trial and error, but I managed to figure out a good test case for the issue we discussed before, @Wadeck.

src/main/resources/lib/credentials/select/select.js
```diff
@@ -260,6 +260,17 @@ Behaviour.specify("INPUT.credentials-select", 'credentials-select', -100, functi
   }).bind($(x));
   x.onchange();
 });
+Behaviour.specify("DIV.include-user-credentials", 'include-user-credentials', 0, function (e) {
+    e.querySelector("input[name='includeUser']").onclick = function (evt) {
+        var caution = e.querySelector('span.user-credentials-caution');
+        caution.hidden = !this.checked;
```
Wadeck: FTR the hidden attribute is valid under IE11, Chrome (since 6), Firefox (since 4).

src/main/resources/lib/credentials/select.jelly
```diff
@@ -75,6 +80,29 @@
...
+            <span class="warning user-credentials-caution" hidden="hidden">
+              ${%userCredentialsCaution}
+            </span>
+            <a href="#" class="help-btn"><l:icon class="icon-help icon-sm" alt="${%Help}"/></a>
+            <div class="help-content" hidden="hidden">
+              <p>${%includeUserCredentialsHelp}</p>
+              <p class="from-plugin">
+                (from <a href="https://wiki.jenkins.io/display/JENKINS/Credentials+Plugin">Credentials Plugin</a>)
+              </p>
+            </div>
```
Wadeck: The `p` should be `div` to avoid adding too much space in the help panel. Otherwise the current style is very accurately similar to the rest of the application, good job!

<details>

<summary>Screenshot comparison</summary>

With `p`:
Screenshot_2019-08-26_091746_001

With `div`:
Screenshot_2019-08-26_091756_001

</details>

It's interesting to notice that 90% of the other help are using `div` but there are still some using `p` which look weird in comparison.

Tested with Chrome, Firefox, IE11, Edge: it works fine.

PR #120, opened 2019-06-21
Show credential ID in DomainWrapper view

chriskilding:
Currently the credential ID is shown in the global credentials screen (`/credentials`), but not in the domain-scoped credentials screen (`/credentials/store/system/domain/_/`), which makes it a little harder for users to work with domain-scoped credentials.

This PR adds the credential ID to the domain-scoped credentials screen, so that users no longer need to look at the path in their address bar to work out the credential ID they should paste into their Jenkinsfiles.

### DIFF
--- a/src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/DomainWrapper/index.jelly
+++ b/src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/DomainWrapper/index.jelly
@@ -39,6 +39,9 @@
             <st:nbsp/>
           </th>
           <th>
+            ${%ID}
+          </th>
+          <th>
             ${%Name}
           </th>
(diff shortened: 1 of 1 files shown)


chriskilding:
Note: the proposed design of the domain-scoped credentials screen is not the final one, hence this PR is marked as a Draft to let us discuss it.

chriskilding:
This is what the proposed change would render:

image001

jeffret-b:
Makes sense to me. I don't see any reason the ID isn't shown for domains. I'm not great at UI design, but this proposal looks simple and satisfactory to me.

jvz:
See linked PR for acceptance test updates affected by this change. It seems they were a bit fragile and were checking for the description on the tooltip! I've updated the test to do an additional xpath query for the updated UI.

PR #121, opened 2019-07-10
Fix Javadoc Error using @link byte[]

batmat:
Using `{@link byte[]}` on JDK11 crashes, cf. https://bugs.openjdk.java.net/browse/JDK-8200432

So we change this construct to move forward. (To unblock https://github.com/jenkinsci/credentials-plugin/pull/119)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.43</version>
+    <version>3.47</version>
     <relativePath />
   </parent>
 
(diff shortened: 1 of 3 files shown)


batmat:
Note: first commit is only bumping the parent pom. So a **crash is still expected**.

Goal is to:
1. See if the `master` branch should actually fail (and didn't until now because, probably, something changed and `master` wasn't rebuilt since)
2. Make the fix here, so https://github.com/jenkinsci/credentials-plugin/pull/119 can be moved forward then cc @jvz

batmat:
@jenkinsci/java11-support for more eyes/review and information (as this is a Java 11-only failure).

batmat:
Thanks everyone!

@oleg-nenashev @stephenc as maintainers (at least releasers fwict :)), could you please merge this? Thanks!

PR #123, opened 2019-08-13
Replace deprecated API use of Jenkins' getInstance() and getActiveIntance()

darxriggs:


jvz:
Built locally, tests pass. Unsure why CI failed. Will review in a little while.

darxriggs:
I accidentally broke a Javadoc. It's fixed now.
On ci.jenkis.io there are more Maven targets run than on a local `mvn verify`. That's why locally everything is fine.

src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java
```diff
@@ -221,12 +221,10 @@ public final FormValidation doCheckId(@ContextInPath ModelObject context, @Query
             }
             if (!(context instanceof Jenkins)) {
                 // CredentialsProvider.lookupStores(User) does not return SystemCredentialsProvider.
-                Jenkins j = Jenkins.getInstance();
-                if (j != null) {
-                    problem = checkForDuplicates(value, context, j);
-                    if (problem != null) {
-                        return problem;
-                    }
+                Jenkins j = Jenkins.get();
```
jvz: This is technically a change in behavior. Before, it would be a no-op if it returns null. Now this will throw an exception.

darxriggs: Exactly. The method `doCheckId` is about form validation. Therefore it should run on Jenkins master and Jenkins should be up and running.

jvz: Ah, then there really was no point to not throwing an exception initially. Good catch!

jvz:
Ok, figured out how to test this in Java 11. Had to set `jenkins.version` to something more recent that supports Java 11. Ran locally and tests still pass.

darxriggs:
The `buildPlugin.recommendedConfigurations()` in the `Jenkinsfile` does this automatically. So on ci.jenkins.io tests are run with Java 8 & 11.

PR #132, opened 2020-01-13
bump parent pom

casz:
blocking #133

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.47</version>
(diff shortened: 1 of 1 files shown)


casz:
all greenio :)

pom.xml
```diff
@@ -95,6 +95,12 @@
         <scope>import</scope>
         <type>pom</type>
       </dependency>
+      <dependency>
+        <groupId>org.hamcrest</groupId>
+        <artifactId>hamcrest-core</artifactId>
+        <version>2.2</version>
```
jglick: Something is wrong here, this is supposed to be pinned by the parent POM. I suspect some other dependency is throwing things off. Oh well, can be fixed up some other time.

PR #134, opened 2020-02-07
Migrate CHANGELOG from wiki to GitHub

MarkEWaite:


README.md
```diff
@@ -43,6 +43,11 @@ To install:
 	mvn release:prepare release:perform -B
 
 
+### Plugin changelog
```
oleg-nenashev: It is not the readme which is used on the plugin site. https://github.com/jenkinsci/credentials-plugin/blob/master/docs/README.adoc also needs to be updated

PR #136, opened 2020-02-24
Add the possibility to disable fingerprintings

Wadeck:
After some reports about millions of fingerprints generated but without the capacity to reproduce the problem, I went at least with this proposal: Adding the possibility to disable the fingerprints.

<details>

<summary>Screenshots</summary>

### Configuration page with help
Screenshot_2020-02-24_171013_001

⚠️ Was replaced by the system property ⚠️ 

### In credentials, when enabled (default)
Screenshot_2020-02-24_170639_001

### In credentials, when disabled
Screenshot_2020-02-24_170529_001

</details>

To disable the fingerprint, you have to 
- either set the system property 
`-Dcom.cloudbees.plugins.credentials.CredentialsProvider.fingerprintEnabled=false` 
- or inside Groovy Script:
`com.cloudbees.plugins.credentials.CredentialsProvider.FINGERPRINT_ENABLED=false`

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
@@ -224,7 +224,7 @@ public abstract class CredentialsProvider extends Descriptor<CredentialsProvider
      */
     @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "Accessible via System Groovy Scripts")
     @Restricted(NoExternalUse.class)
-    public static /* not final */ boolean FINGERPRINT_ENABLED = SystemProperties.getBoolean(FINGERPRINT_ENABLED_NAME, true);
+    public static /* not final */ boolean FINGERPRINT_ENABLED = Boolean.parseBoolean(System.getProperty(FINGERPRINT_ENABLED_NAME, "true"));
     
     /**
      * Default constructor.
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
@@ -1462,14 +1462,18 @@ public abstract class CredentialsProvider extends Descriptor<CredentialsProvider
      */
     @NonNull
     public static <C extends Credentials> List<C> trackAll(@NonNull Run build, @NonNull List<C> credentials) {
-        for (Credentials c : credentials) {
-            if (c != null) {
-                try {
-                    getOrCreateFingerprintOf(c).addFor(build);
-                } catch (IOException e) {
-                    LOGGER.log(Level.FINEST, "Could not track usage of " + c, e);
+        if (CredentialsProviderManager.isFingerprintEnabledOrDefault()) {
+            for (Credentials c : credentials) {
+                if (c != null) {
+                    try {
+                        getOrCreateFingerprintOf(c).addFor(build);
+                    } catch (IOException e) {
+                        LOGGER.log(Level.FINEST, "Could not track usage of " + c, e);
+                    }
                 }
             }
+        } else {
+            LOGGER.log(Level.FINEST, "TrackAll(Run) called but fingerprints disabled in the global configuration.");
         }
         return credentials;
     }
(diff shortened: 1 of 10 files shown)


src/main/resources/com/cloudbees/plugins/credentials/CredentialsProviderManager/Configuration/help-fingerprintEnabled.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    The fingerprints of credentials are used to track their usage inside the application.
+    If you disable them, you will no longer be able to know where the credentials is used.
```
jvz:
```suggestion
    If you disable them, you will no longer be able to know where the credentials are used.
```

src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/CredentialsWrapper/index.properties
```diff
@@ -22,3 +22,5 @@
 # THE SOFTWARE.
 #
 usageTrackingImperfect=Note: usage tracking requires the cooperation of plugins and consequently may not track every use.
+fingerprintDisabledWarning=The fingerprints are currently disabled and they are required to track the usage. \
```
jvz:
```suggestion
fingerprintDisabledWarning=Credential fingerprints are currently disabled which is required to track their usage. \
```

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -1462,14 +1462,18 @@ public static Fingerprint getOrCreateFingerprintOf(@NonNull Credentials c) throw
...
+                if (c != null) {
+                    try {
+                        getOrCreateFingerprintOf(c).addFor(build);
+                    } catch (IOException e) {
+                        LOGGER.log(Level.FINEST, "Could not track usage of " + c, e);
+                    }
                 }
             }
+        } else {
+            LOGGER.log(Level.FINEST, "TrackAll(Run) called but fingerprints disabled in the global configuration.");
```
jeffret-b: It's a little odd that one parameter type is mentioned but the other isn't.

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -1519,51 +1523,55 @@ public static Fingerprint getOrCreateFingerprintOf(@NonNull Credentials c) throw
...
-                        change.commit();
+                    } catch (IOException e) {
+                        LOGGER.log(Level.FINEST, "Could not track usage of " + c, e);
                     }
-                } catch (IOException e) {
-                    LOGGER.log(Level.FINEST, "Could not track usage of " + c, e);
                 }
             }
+        } else {
+            LOGGER.log(Level.FINEST, "TrackAll(Node) called but fingerprints disabled in the global configuration.");
```
jeffret-b: I suppose the intention is to distinguish between the node and run methods, but it feels a little confusing to have part of the signature. (Not sure it's really intended to show the signature, but that's kind of what it looks like.)

src/main/java/com/cloudbees/plugins/credentials/CredentialsProviderManager.java
```diff
@@ -137,6 +146,17 @@ public static CredentialsProviderManager getInstanceOrDie() {
         return instance;
     }
 
+    /**
+     * Helper method to know if the fingerprints are processed
+     *
+     * @return if the fingerprints are processed
+     * @since TODO
+     */
+    public static boolean isFingerprintEnabledOrDefault() {
```
jeffret-b: Seems a little longer than needed. Especially the OrDefault. That probably folds into the enabled case.

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
Wadeck: The intent was to distinguish between the node, the item and the run versions. I will change the format to avoid confusion with signature.

Wadeck:
> Are there any other places in the system that could be impacted by these changes?

Do you mean the call of `getOrCreateFingerprintOf` outside of the plugin? Nothing according to https://github.com/search?q=org%3Ajenkinsci+getOrCreateFingerprintOf&type=Code. The plugin provides three ways to track the usage, for Node, Run and Item. These methods are used and they are covered.

----

Moved to system property instead of UI setting. Jeff was for that and I was neutral. If you have opinion, voice it ;) We can always add the UI setting back in the future if we get feedbacks.

src/main/java/com/cloudbees/plugins/credentials/CredentialsProviderManager.java
Wadeck: Method removed due to the system property 👍

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -209,6 +212,20 @@
...
+    public static final String FINGERPRINT_ENABLED_NAME = CredentialsProvider.class.getSimpleName() + ".fingerprintEnabled";
+    
+    /**
+     * Control if the fingerprints must be used or not. 
+     * By default they are activated and thus allow the tracking of credentials usage.
+     * In case of performance troubles in some weird situation, you can disable the behavior by setting it to {@code false}.
+     */
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "Accessible via System Groovy Scripts")
+    @Restricted(NoExternalUse.class)
+    public static /* not final */ boolean FINGERPRINT_ENABLED = SystemProperties.getBoolean(FINGERPRINT_ENABLED_NAME, true);
```
jeffret-b: There is some discussion elsewhere on whether this form should be used.

Wadeck: Please provide a link to the related discussion

daniel-beck:
Given that this appears to be a workaround for a rare and unreproducible problem, it seems overkill to add a UI option for it…?

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
daniel-beck: So far the only counter-argument I've seen is use from `jenkins-test-harness`.

Wadeck:
> Given that this appears to be a workaround for a rare and unreproducible problem, it seems overkill to add a UI option for it…?

In that case you're fine with the current system property right? ^^

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
jeffret-b: See [discussion](https://github.com/jenkinsci/jenkins/pull/4389#discussion_r369716605).

daniel-beck:
@Wadeck Yes. (Do I misunderstand what this PR is doing? Haven't looked at the code yet.)

src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/CredentialsWrapper/index.properties
```diff
@@ -22,3 +22,5 @@
 # THE SOFTWARE.
 #
 usageTrackingImperfect=Note: usage tracking requires the cooperation of plugins and consequently may not track every use.
+fingerprintDisabledWarning=Credential fingerprints are currently disabled which is required to track their usage. \
```
jeffret-b: There are a couple of problems with this "which" clause.
I suggest something simple like this.
```suggestion
fingerprintDisabledWarning=Credential fingerprints are disabled. \
```

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -209,6 +212,20 @@
...
+    public static final String FINGERPRINT_ENABLED_NAME = CredentialsProvider.class.getSimpleName() + ".fingerprintEnabled";
+    
+    /**
+     * Control if the fingerprints must be used or not. 
+     * By default they are activated and thus allow the tracking of credentials usage.
+     * In case of performance troubles in some weird situation, you can disable the behavior by setting it to {@code false}.
+     */
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "Accessible via System Groovy Scripts")
+    @Restricted(NoExternalUse.class)
+    public static /* not final */ boolean FINGERPRINT_ENABLED = Boolean.parseBoolean(System.getProperty(FINGERPRINT_ENABLED_NAME, "true"));
```
jeffret-b: While I find it to be a minority opinion, I think James makes a valid point, so this should not be `public`.

jvz:
Rebuilding

jvz:
Released in 2.3.5. Thanks for the PR!

PR #138, opened 2020-02-28
Use working link to CHANGELOG

MarkEWaite:
Also remove reference to wiki documentation.  Wiki link immediately redirects to plugins.jenkins.io which is this file.


jvz:
Thanks for the PR!

PR #139, opened 2020-03-05
Add support for future Manage Jenkins layout

daniel-beck:
Downstream from https://github.com/jenkinsci/jenkins/pull/4546


daniel-beck:
That… seems a bit premature.

jvz:
It's harmless code, and I haven't released it yet. :)

daniel-beck:
@jvz Could you adapt this plugin to the changed upstream code?

jvz:
> @jvz Could you adapt this plugin to the changed upstream code?

Now it's strongly typed? How can I support that without requiring a recent Jenkins version?

daniel-beck:
@jvz Rename to `getCategoryName`, which is still a string and remains for compatibility.

jvz:
See #140

PR #144, opened 2020-04-13
[JENKINS-61880] GlobalCredentialsConfiguration JCasC compatibility

amuniz:
See [JENKINS-61880](https://issues.jenkins-ci.org/browse/JENKINS-61880)

Add a `@Symbol` to the category so it's rendered properly in JCasC YAML.

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/GlobalCredentialsConfiguration.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/GlobalCredentialsConfiguration.java
@@ -42,6 +42,7 @@ import javax.servlet.ServletException;
 import jenkins.model.GlobalConfigurationCategory;
 import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
+import org.jenkinsci.Symbol;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.HttpResponse;
@@ -221,6 +222,7 @@ public class GlobalCredentialsConfiguration extends ManagementLink
      * Security related configurations.
      */
     @Extension
+    @Symbol("globalCredentialsConfiguration")
     public static class Category extends GlobalConfigurationCategory {
         /**
          * {@inheritDoc}
--- a/src/test/java/com/cloudbees/plugins/credentials/casc/CredentialsCategoryTest.java
+++ b/src/test/java/com/cloudbees/plugins/credentials/casc/CredentialsCategoryTest.java
@@ -1,0 +1,57 @@
+package com.cloudbees.plugins.credentials.casc;
+
+import com.cloudbees.plugins.credentials.GlobalCredentialsConfiguration;
+import hudson.Extension;
+import hudson.ExtensionList;
+import io.jenkins.plugins.casc.misc.ConfiguredWithCode;
+import io.jenkins.plugins.casc.misc.JenkinsConfiguredWithCodeRule;
+import jenkins.model.GlobalConfiguration;
+import jenkins.model.GlobalConfigurationCategory;
+import org.jenkinsci.Symbol;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.TestExtension;
+
+import javax.annotation.Nonnull;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class CredentialsCategoryTest {
+
+    @Rule
+    public JenkinsConfiguredWithCodeRule r = new JenkinsConfiguredWithCodeRule();
+
+    @Test
+    @ConfiguredWithCode("credentials-category.yaml")
+    public void globalCredentialsConfigurationCategory() {
+        assertThat(ExtensionList.lookupSingleton(TestGlobalConfiguration.class).getConfig(), equalTo("hello"));
+    }
+
+    @TestExtension
+    @Symbol("testGlobalConfiguration")
+    public static class TestGlobalConfiguration extends GlobalConfiguration {
+
+        private String config;
+
+        @Nonnull
+        @Override
+        public GlobalConfigurationCategory getCategory() {
+            return GlobalConfigurationCategory.get(GlobalCredentialsConfiguration.Category.class);
+        }
+
+        @Nonnull
+        @Override
+        public String getDisplayName() {
+            return "Test Global Configuration";
+        }
+
+        public String getConfig() {
+            return config;
+        }
+
+        public void setConfig(String config) {
+            this.config = config;
+        }
+    }
+}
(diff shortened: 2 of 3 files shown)


src/main/java/com/cloudbees/plugins/credentials/GlobalCredentialsConfiguration.java
```diff
@@ -221,6 +222,7 @@ public String getDisplayName() {
      * Security related configurations.
      */
     @Extension
+    @Symbol("globalCredentialsConfiguration")
```
timja, line 225: shouldn't it just be `globalCredentials` (unless there's a clash)?
```suggestion
    @Symbol("globalCredentials")
```

what does it look like in the export currently?

amuniz: Like this:

```
globalCredentialsCategory:
  cyberArkGlobalConfiguration:
    allowCustomAppIds: false
```

Before:

```
? ''
: cyberArkGlobalConfiguration:
    allowCustomAppIds: false
```

amuniz:
After filing it, I'm thinking this is a backward breaking change... :thinking: 

The feature was not broken, it just rendered badly... but worked fine, so with this change the category has now a named root in the yaml so the previous ugly format does not work anymore.

amuniz:
However, AFAICT it's not really used anywhere else: https://github.com/search?q=org%3Ajenkinsci+GlobalCredentialsConfiguration&type=Code

(I have a non-public implementation using it, that's why I found the issue).

src/main/java/com/cloudbees/plugins/credentials/GlobalCredentialsConfiguration.java
timja: I don't really understand why this isn't in the 'credentials' category (but possibly too late).

but the symbol shouldn't have a suffix of category as it's implied by the extension point.

Do you have a symbol added for `cyberArkGlobalConfiguration`? normally the `GlobalConfiguration` would be stripped out by jcasc unless you actually put it in your symbol name.

I would expect the yaml to be:

```yaml
globalCredentials:
  cyberArk:
    allowCustomAppIds: false
```

amuniz: > I don't really understand why this isn't in the 'credentials' category (but possibly too late).

Because it's not defining _credentials_, it's defining credential stores. It's under `/jenkins/configureCredentials/` - I didn't even know about this feature! Just learnt it today.

amuniz: > Do you have a symbol added for cyberArkGlobalConfiguration? normally the GlobalConfiguration would be stripped out by jcasc unless you actually put it in your symbol name.

Yeah, I have a symbol, but it doesn't really have any impact on this, I can set the symbol there to whatever and the root element is still empty (because it's the category!) - I'm not saying this makes sense, just stating the current behavior.

So it looks like `GlobalCategory` extensions behave differently when they are in the `GlobalCredentialsConfiguration.Category` (which is the global configuration implementing [`Descriptor#getCategory`](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/Descriptor.java#L831), which from a descriptor configuration stand point kind of makes sense as [`GlobalCredentialsConfiguration#configure`](https://github.com/jenkinsci/credentials-plugin/blob/00173e541b29a5c742954ecd69f509d35ca92208/src/main/java/com/cloudbees/plugins/credentials/GlobalCredentialsConfiguration.java#L163) is acting as a root configurator element for all `GlobalConfiguration` under the category.

timja:
> After filing it, I'm thinking this is a backward breaking change... 🤔
> 
> The feature was not broken, it just rendered badly... but worked fine, so with this change the category has now a named root in the yaml so the previous ugly format does not work anymore.

It was broken afaik, and not round trippable, we've came across this before and there was a hack put into JCasc because of it: https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/plugin/src/main/java/io/jenkins/plugins/casc/impl/configurators/GlobalConfigurationCategoryConfigurator.java#L125

This should fix the actual issue though and we should be able to remove the hack

amuniz:
Thanks for the review @timja!

About the Symbol name, `globalCredentials` seems confusing to me. It's not really defining _Global Credentials_ but _Global Configuration for Credentials_ (additional credentials stores actually). So I think having `configuration` in the Symbol name is making it more obvious.

src/main/java/com/cloudbees/plugins/credentials/GlobalCredentialsConfiguration.java
jetersen: I applied a patch against discovering the empty category.
https://github.com/jenkinsci/configuration-as-code-plugin/blob/02ca4728571a26465339742f532c9e50fe629b07/plugin/src/main/java/io/jenkins/plugins/casc/impl/configurators/GlobalConfigurationCategoryConfigurator.java#L125-L129

jetersen: We also have code that removes `Category` 
https://github.com/jenkinsci/configuration-as-code-plugin/blob/02ca4728571a26465339742f532c9e50fe629b07/plugin/src/main/java/io/jenkins/plugins/casc/impl/configurators/GlobalConfigurationCategoryConfigurator.java#L58-L62

timja: This should allow removing that patch? The second example is for guessing the name and is the root cause of this

jetersen: Well we could remove it if it were not for older versions of the plugin.

Does it even work with this patch @amuniz when exporting it?

timja: https://github.com/jenkinsci/credentials-plugin/pull/144#discussion_r407452543

amuniz: > Please use a recently new version of JCasC

I'm using the latest release (1.38) in my local tests.

> Does it even work with this patch @amuniz when exporting it?

Yes, it does.

jetersen: I guess it works because the child is discovered 🤔

amuniz: I just added a test to demonstrate the behavior, it might be useful to understand the issue @jetersen

jvz:
Is this ready for review?

amuniz:
Yes.

amuniz:
Thanks for your quick response @jvz!

jvz:
Thanks for the PR!

PR #150, opened 2020-05-25
Move 'Manage Credentials' link to Manage Jenkins for Administrators

timja:
https://www.jenkins.io/events/online-hackfest/2020-uiux/
'Credentials Management UX'

There will need to be a separate change to Jenkins core to allow showing the Manage Jenkins page with another permission, I'll take a look at that as well, but it shouldn't block this change

image

cc @jvz


src/main/resources/com/cloudbees/plugins/credentials/Messages.properties
```diff
@@ -71,7 +71,7 @@ SystemCredentialsProvider.ProviderImpl.DisplayName=Jenkins Credentials Provider
 SystemCredentialsProvider.UserFacingAction.DisplayName=System
 UserCredentialsProvider_DisplayName=Credentials
 UserCredentialsProvider.UserFacingAction.DisplayName=User
-GlobalCredentialsConfiguration.DisplayName=Configure Credentials
+GlobalCredentialsConfiguration.DisplayName=Configure Credentials Providers
```
res0nance, line 74:
```suggestion
GlobalCredentialsConfiguration.DisplayName=Configure Credential Providers
```

oleg-nenashev:
Should we also remove it from the Admin's sidepanel while we are here? Can be a separate PR

src/main/resources/com/cloudbees/plugins/credentials/Messages.properties
timja: Not sure, They are called 'Credentials' everywhere

timja:
> Should we also remove it from the Admin's sidepanel while we are here? Can be a separate PR

I would like to do it in a separate PR, I do plan to do it though

src/main/resources/com/cloudbees/plugins/credentials/Messages.properties
res0nance: I mentioned this because the subtext in the screenshot shows `Credential Providers`

timja:
> Should we also remove it from the Admin's sidepanel while we are here? Can be a separate PR

PR to allow anyone to be able to view credentials from 'Manage Jenkins' https://github.com/jenkinsci/jenkins/pull/4738

timja:
> Should we also remove it from the Admin's sidepanel while we are here? Can be a separate PR

Done in https://github.com/jenkinsci/credentials-plugin/pull/150/commits/b2de99c0df2f8b775d9d418f0a6ddbc6fb36603b

src/main/java/com/cloudbees/plugins/credentials/ManageCredentialsConfiguration.java
```diff
@@ -0,0 +1,43 @@
+package com.cloudbees.plugins.credentials;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.Extension;
+import hudson.model.ManagementLink;
+import hudson.security.Permission;
+
+/**
+ * {@link ManagementLink} to expose a link to the manage credentials configuration screen.
+ *
```
oleg-nenashev, line 10:
```suggestion
 * 
 * @since TODO
```

src/main/java/com/cloudbees/plugins/credentials/ManageCredentialsConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+@Extension(ordinal = Integer.MAX_VALUE - 211)
```
oleg-nenashev, line 12:
```suggestion
@Extension(ordinal = Integer.MAX_VALUE - 211)
@Restricted(NoExternalUse.class)
```

res0nance:
We definitely need to highlight the Action Change in the release notes. So admins will not be surprised by this change

jvz:
Looks like a neat idea. Will the [ATH](https://github.com/jenkinsci/acceptance-test-harness) need to be updated, too?

timja:
> Looks like a neat idea. Will the [ATH](https://github.com/jenkinsci/acceptance-test-harness) need to be updated, too?

From what I can tell everything goes to the URL directly nothing clicks the link, i.e. https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/CredentialsPage.java#L24

It would be good if the credentials ATH tests were run on PR though...

(I haven't run the tests)

maemile:
Hello,
This PR impact also credentials link at folder level (with folders plugin). So an admin as to type /credentials at folder level to manage it. 
Is that expected or a regression?
Thanks :)

jvz:
That sounds odd. What was the previous behavior?

maemile:
> That sounds odd. What was the previous behavior?

The credentials link was available at folder level for an admin.

You can check the behavior in the doc:
https://docs.cloudbees.com/docs/admin-resources/latest/plugins/folder#folder-sect-sample

jvz:
Ah, this is a regression. Can you file a Jira issue? https://issues.jenkins-ci.org/

Vlatombe:
There is also a regression for credentials defined at user level. See #155

src/main/java/com/cloudbees/plugins/credentials/ViewCredentialsAction.java
```diff
@@ -223,11 +223,14 @@ public String getStoreBaseUrl(String itUrl) {
...
     public boolean isVisible() {
-        if (context instanceof AccessControlled
-                && !((AccessControlled) context).hasPermission(CredentialsProvider.VIEW)) {
-            // must have permission
-            return false;
+        if (context instanceof AccessControlled) {
+            AccessControlled accessControlled = (AccessControlled) this.context;
+            if (accessControlled.hasPermission(Jenkins.ADMINISTER) || !accessControlled.hasPermission(CredentialsProvider.VIEW)) {
+                // must have permission, Administrator's view credentials from 'Manage Jenkins'
+                return false;
```
jglick, line 230: FWIW I found this (b2de99c0df2f8b775d9d418f0a6ddbc6fb36603b) very confusing and reported it as a bug for CloudBees CI until I tracked this down in history: by _gaining_ the Overall/Administer permission, a sidebar link is suddenly hidden from you, which is the opposite of what you would normally expect.

PR #154, opened 2020-06-18
Fix accessing folder credentials

timja:


SMillerDev:
Could this be released as well? I now have a bunch of confused colleagues since the last update.

jvz:
Oops, looks like I missed this one. I'll make a release later today, yes.

jvz:
Pushed to 2.3.10.

PR #155, opened 2020-06-29
Credentials action should be hidden only for the root Jenkins, not for folders or users

Vlatombe:
Fix regression from #150, amends #154 

/cc @timja


Vlatombe:
Not sure what is going on at the CI level though...

timja:
> Not sure what is going on at the CI level though...

that read time out is a flake I've seen before, I retriggered when I approved it =/

Vlatombe:
also /cc @jvz

timja:
Not sure what's going on, re-triggered it a few times, and ran the tests locally on java 11 (all pass).

PR #156, opened 2020-06-30
Update Jenkins to 2.222.4

jvz:
This also updates the plugin pom and removes some code no longer
necessary thanks to the upgrade.


src/main/java/com/cloudbees/plugins/credentials/GlobalCredentialsConfiguration.java
```diff
@@ -164,7 +165,7 @@ private boolean configure(StaplerRequest req, JSONObject json) throws FormExcept
 
         // persist all the provider configs
         boolean result = true;
-        for (Descriptor<?> d : Functions.getSortedDescriptorsForGlobalConfig(FILTER)) {
+        for (Descriptor<?> d : Functions.getSortedDescriptorsForGlobalConfigByDescriptor(Category.class::isInstance)) {
```
jglick, line 168: does not compile

jglick: you need to also

```diff
diff --git Jenkinsfile Jenkinsfile
index 87a086d..a229fa5 100644
--- Jenkinsfile
+++ Jenkinsfile
@@ -1 +1 @@
-buildPlugin(configurations: buildPlugin.recommendedConfigurations())
+buildPlugin()
```

jvz: It compiles. Is it not supposed to?

jglick: The current `Jenkinsfile` tries to build the plugin against an _older_ version of Jenkins, which will not work.

jglick: If you have write permission to the plugin, you can make the above change. Otherwise you need to enlist the help of someone who does.

src/main/java/com/cloudbees/plugins/credentials/GlobalCredentialsConfiguration.java
```diff
@@ -68,6 +68,7 @@
     /**
      * Our filter.
      */
+    @Deprecated
```
jglick, line 71: Any particular reason this is being deprecated?

jvz: That's an old constant using the Guava `Predicate` interface which was deprecated (and marked with a `@DoNotUse` which failed the build otherwise). The update just uses an inline SAM to align to the new method that uses the `java.util.function` version.

jvz: The deprecated code was the `Functions` method by the way, not Guava itself.

jglick: …because you need a `Predicate<Descriptor>` now.

jvz: The field is public. I didn't want to remove it right away, and the field is useless ever since I refactored it to use a lambda instead of an anonymous class anyways.

jvz: Oh I see what you mean.

src/main/java/com/cloudbees/plugins/credentials/GlobalCredentialsConfiguration.java
jvz: Fixed in https://github.com/jenkinsci/credentials-plugin/pull/157

PR #157, opened 2020-06-30
Fix runtime error from type erasure

jvz:
Oops.

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/GlobalCredentialsConfiguration.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/GlobalCredentialsConfiguration.java
@@ -23,7 +23,6 @@
  */
 package com.cloudbees.plugins.credentials;
 
-import com.google.common.base.Predicate;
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/plugins/credentials/GlobalCredentialsConfiguration.java
```diff
@@ -68,8 +68,8 @@
     /**
      * Our filter.
      */
-    @Deprecated
-    public static final Predicate<GlobalConfigurationCategory> FILTER = Category.class::isInstance;
+    @SuppressWarnings("rawtypes")
+    public static final Predicate<Descriptor> FILTER = d -> d.getCategory() instanceof Category;
```
jglick, line 72: Potentially incompatible. Is any other plugin using this field?

jvz: I did a search using usage-in-plugins and found nothing.

PR #164, opened 2020-07-31
Simplified use of User.get and ACL.as

jglick:
Not sure what 5ad742bac2d503c1b0537799985a50af1f5b1885 was about, but this is simpler and uses currently recommended APIs better.


jvz:
CI failure looks unrelated (something about incrementals not working).

PR #166, opened 2020-11-06
Add more documentation along with minor maintenance

jvz:
This PR has a few items:

* Updated parent pom
* SafeVarargs annotations on some things
* Added documentation page about feature flags (system properties)
* Started an FAQ with some higher level concerns based on frequent user confusion (and a little bit of inspiration from recent research done regarding external secrets managers)

### DIFF
--- a/docs/faq.adoc
+++ b/docs/faq.adoc
@@ -19,12 +19,11 @@ Examples of common external credentials stores for cloud providers or self-hoste
 
 A full listing (including some developer-level details) of https://www.jenkins.io/doc/developer/extensions/credentials/[plugins that provide credentials extensions] is available in the developer documentation.
 
-=== How should I organize my credentials?
+=== How do credentials get accessed?
 
-It's important to understand the access model of credentials in order to organize them most appropriately.
 Credentials are accessed through a combination of scopes and contexts, which are further categorized into types and domains.
 
-The _context_ of a credential is the path in Jenkins where the credentials are accessed.
+The _context_ of a credential is the path in Jenkins (i.e., the URL) where the credentials are accessed.
 Contexts include the root Jenkins context, jobs, users, agents, views, folders, and generally any other kind of model object.
 
 The _scope_ of a credential can be _global, system,_ or _user,_ which determines the visibility of a credential to a context.
@@ -37,18 +36,20 @@ Global-scoped credentials are visible to their associated contexts and any child
 Global-scoped credentials in the Jenkins Credentials Provider (system credentials) are accessible to any job in the system!
 ====
 
-Credentials _stores_ are the underlying storage mechanism for secrets, and credentials _providers_ combine them with the functionality to manage secrets in a context for a certain set of scopes.
+The _domain_ of a credential is a property of credentials that allows them to be grouped together more logically in the UI.
+This does *not* provide any security feature itself and is strictly enforced through the UI.
+
+=== How should I organize my credentials?
+
+Different credentials _providers_ give different means of organizing and exposing credentials.
 For example, the Jenkins Credentials Provider uses a system-wide credentials store at the root context and supports both system and global scoped credentials.
 Folder plugins provide a Folder Credentials Provider to anchor credentials storage in a folder along with the advanced authorization strategies possible there.
 User-scoped credentials are primarily accessed through the User Credentials Provider which stores secrets with the user profile in Jenkins.
 
-Credentials _types_ include things like username/password, secret text, SSH key, etc.
-Finally, credentials _domains_ provide a way to help organize credentials by domain name, URI, protocol, and other requirements which can make it much easier to identify the appropriate credentials to use from the UI.
-(Domains are _not_ enforced at usage time, only for purposes of the UI.)
-
 Credentials should be organized to minimize granting access.
-Common conventions include splitting up credentials by team or product into separate folders.
+This can make use of both logical organization into team or product folders as well as fine grained ACLs to grant access to credentials per principal.
 Deployment credentials can be added to a user's profile to be used interactively in a deployment job to further isolate credentials access.
+Enabling the _Credentials/UseOwn_ or _Credentials/UseItem_ permissions (link:fflags.adoc[feature flags documentation]) allows for further refinement of permissions (e.g., to allow some users to run a build, but only allow certain users to use its folder's or triggering user's deployment credentials).
 
 === What plugins do I need to be secure?
 
--- a/docs/faq.adoc
+++ b/docs/faq.adoc
@@ -1,4 +1,4 @@
-== FAQs
+== FAQ
 
 This document provides a quick reference for frequently questioned concepts and features in Credentials.
 
--- a/docs/faq.adoc
+++ b/docs/faq.adoc
@@ -7,7 +7,7 @@ This document provides a quick reference for frequently questioned concepts and 
 By default, Jenkins uses an internal credentials store where credentials secret data are encrypted at rest in the Jenkins controller file system.
 This security model does not prevent anyone with read access to the controller file system from reading and decrypting those secrets.
 For more sensitive secrets, it is highly recommended to install and use an external credentials provider plugin which delegates the secure storage of secrets to a proper key management system.
-Various cloud providers have secrets management and HSMs available for this purpose, and open source software such as Vault or Conjur provide more generic secrets management for multiple platforms.
+Various cloud providers (including https://plugins.jenkins.io/aws-secrets-manager-credentials-provider/[AWS], https://plugins.jenkins.io/azure-keyvault/[Azure], and https://plugins.jenkins.io/gcp-secrets-manager-credentials-provider/[GCP]) have secrets management and HSMs available for this purpose, and open source software such as https://plugins.jenkins.io/kubernetes-credentials-provider/[Kubernetes], https://plugins.jenkins.io/hashicorp-vault-plugin/[Vault], or https://plugins.jenkins.io/conjur-credentials/[Conjur] provide more generic secrets management for multiple platforms.
 
 === How should I organize my credentials?
 
@@ -24,7 +24,7 @@ Global-scoped credentials are visible to their associated contexts and any child
 
 [IMPORTANT]
 ====
-Global scoped credentials in the Jenkins Credentials Provider (system credentials) are accessible to any job in the system!
+Global-scoped credentials in the Jenkins Credentials Provider (system credentials) are accessible to any job in the system!
 ====
 
 Credentials _stores_ are the underlying storage mechanism for secrets, and credentials _providers_ combine them with the functionality to manage secrets in a context for a certain set of scopes.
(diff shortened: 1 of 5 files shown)


docs/faq.adoc
```diff
@@ -0,0 +1,58 @@
+== FAQs
+
+This document provides a quick reference for frequently questioned concepts and features in Credentials.
+
+=== Where do credentials get stored?
+
+By default, Jenkins uses an internal credentials store where credentials secret data are encrypted at rest in the Jenkins controller file system.
+This security model does not prevent anyone with read access to the controller file system from reading and decrypting those secrets.
+For more sensitive secrets, it is highly recommended to install and use an external credentials provider plugin which delegates the secure storage of secrets to a proper key management system.
+Various cloud providers have secrets management and HSMs available for this purpose, and open source software such as Vault or Conjur provide more generic secrets management for multiple platforms.
```
jglick, line 10: Maybe mention `kubernetes-credentials-provider` here as well.

docs/faq.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+=== How should I organize my credentials?
+
+It's important to understand the access model of credentials in order to organize them most appropriately.
+Credentials are accessed through a combination of scopes and contexts, which are further categorized into types and domains.
+
+The _context_ of a credential is the path in Jenkins where the credentials are accessed.
+Contexts include the root Jenkins context, jobs, users, agents, views, folders, and generally any other item type.
```
jglick, line 18:
```suggestion
Contexts include the root Jenkins context, jobs, users, agents, views, folders, and generally any other kind of model object.
```

since _item_ refers to folders and jobs but not views or agents for example.

docs/faq.adoc
```diff
@@ -0,0 +1,58 @@
...
+====
+Global scoped credentials in the Jenkins Credentials Provider (system credentials) are accessible to any job in the system!
+====
+
+Credentials _stores_ are the underlying storage mechanism for secrets, and credentials _providers_ combine them with the functionality to manage secrets in a context for a certain set of scopes.
+For example, the Jenkins Credentials Provider uses a system-wide credentials store at the root context and supports both system and global scoped credentials.
+Folder plugins provide a Folder Credentials Provider to anchor credentials storage in a folder along with the advanced authorization strategies possible there.
+User-scoped credentials are primarily accessed through the User Credentials Provider which stores secrets with the user profile in Jenkins.
+
+Credentials _types_ include things like username/password, secret text, SSH key, API token, etc.
```
jglick, line 35: Generally either 
```suggestion
Credentials _types_ include things like username/password, secret text, SSH key, etc.
```

Tokens are generally represented as either passwords with an arbitrary username, or secret text.

docs/faq.adoc
```diff
@@ -0,0 +1,58 @@
...
+Finally, credentials _domains_ provide a way to help organize credentials by domain name, URI, protocol, and other requirements which can make it much easier to identify the appropriate credentials to use from the UI.
```
jglick, line 36:
```suggestion
Finally, credentials _domains_ provide a way to help organize credentials by domain name, URI, protocol, and other requirements which can make it much easier to identify the appropriate credentials to use from the UI.
(Domains are _not_ enforced at usage time, only for purposes of the UI.)
```

docs/faq.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+=== What plugins do I need to be secure?
+
+First, ensure that you are following the https://www.jenkins.io/doc/book/system-administration/security/[Jenkins security best practices].
+In particular, you'll need to use a non-trivial authorization strategy such as https://plugins.jenkins.io/matrix-auth/[matrix], https://plugins.jenkins.io/role-strategy/[role-based], or https://plugins.jenkins.io/folder-auth/[folder-based]; the trivial authorization strategies would defeat the purpose of securing your secrets!
+Next, you'll need the https://plugins.jenkins.io/authorize-project/[Authorize Project] plugin so that more fine-grained permission controls are available to configure to jobs so that credentials accessible to one job aren't necessarily accessible to every other job.
```
jglick, line 42: `authorize-project` is not necessary for that purpose; you can use folder-scoped credentials.

docs/faq.adoc
```diff
@@ -0,0 +1,58 @@
...
+The https://plugins.jenkins.io/ownership/[Job and Node ownership] plugin extends many of the mentioned authorization strategies to allow for further configuration.
+Additional hardening can be applied via the https://plugins.jenkins.io/job-restrictions/[Job Restrictions] plugin.
+
+When working with especially sensitive secrets, use of a hardware security module (HSM) is recommended for physical storage and protection of secret and private keys.
```
jglick, line 46: Not sure how relevant this is in the context of Jenkins credentials.

docs/fflags.adoc
```diff
@@ -0,0 +1,16 @@
+== Feature Flags
+
+The following system properties are available for toggling or specifying particular features to turn on or off.
```
jglick, line 3:
```suggestion
The following system properties are available for toggling or specifying particular features to turn on or off.
Like any system properties, they should be considered experimental and subject to removal without notice.
```

docs/fflags.adoc
```diff
@@ -0,0 +1,16 @@
...
+
+* `com.cloudbees.plugins.credentials.UseOwnPermission` - Enables the Credentials/UseOwn permission type which allows a
```
jglick, line 5: BTW https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

docs/fflags.adoc
```diff
@@ -0,0 +1,16 @@
...
+user to provide their own user-scoped credentials to a build. When this feature flag is not specified, this permission
+is ignored in favor of Job/Build. When the flag is enabled, Credentials/UseOwn is only implied by Overall/Administer by
+default.
+
+* `com.cloudbees.plugins.credentials.UseItemPermission` - Enables the Credentials/UseItem permission type which allows a
```
jglick, line 10: A guide to what “use own” and “use item” are actually _for_ would be appreciated. I for one know they exist, and could with difficulty reverse engineer the source code to determine their effect in some situation, but have no idea how you would leverage them securely to accomplish something.

docs/faq.adoc
jvz: The relevance is that users should not expect HSM-level security with this credentials plugin despite looking super fancy. For example, suppose you want to automate code signing with a TLS certificate. You might want extra hardware security for the signing key so that nothing else can use it. For Jenkins, we use the Azure equivalent of Vault to do that, for example.

docs/fflags.adoc
jvz: You may have noticed me doing that elsewhere in the adoc files ;)

docs/fflags.adoc
jvz: The UseItem permission is fairly simple: by adding this permission, now you can no longer access credentials from the folder or similar when running a build. For UseOwn, that is fairly difficult to describe use cases for without diving into user-scoped credentials in general which are a little anemic still. I do agree, though.

docs/faq.adoc
jglick: Maybe rephrase to clarify that if you want HSM or similar levels of security, you should not be using Jenkins credentials. Currently it makes it sound like there is some vaguely described means of using an HSM to help lock down Jenkins credentials.

jvz: Good idea :)

docs/fflags.adoc
jglick: Nothing involving `authorize-project` is “simple”. :smile:

docs/faq.adoc
jvz: That's true, though folder-stored (there's no folder scope; there's way too many concepts in this plugin) credentials are a little coarser than what `authorize-project` offers. I'll update this to clarify, though, thanks.

docs/fflags.adoc
jvz: Alright, I played around with the feature a little bit and figured out the secret to why it's useful. The main use case is with project-based matrix authorization strategies and similar derivations. Authorize Project can add another dimension to the whole idea, but it's not required for this to be useful. It's a little convoluted, but I think you can basically use the UseItem permission to selectively limit which users can even access credentials that the job (or context) can access itself. If you get that, then you can imagine the flexibility.

docs/fflags.adoc
jvz: Fixed.

docs/faq.adoc
```diff
@@ -0,0 +1,64 @@
+== FAQs
```
daniel-beck, line 1:
```suggestion
== FAQ
```

Per https://en.wikipedia.org/wiki/FAQ

docs/fflags.adoc
```diff
@@ -0,0 +1,23 @@
+== Feature Flags
+
+The following system properties are available for toggling or specifying particular features to turn on or off.
+Like any system properties, they should be considered experimental and subject to removal without notice.
```
daniel-beck, line 4: Perhaps link to https://www.jenkins.io/doc/book/managing/system-properties/ for usage etc.?

docs/fflags.adoc
```diff
@@ -0,0 +1,23 @@
...
+
+* `com.cloudbees.plugins.credentials.UseOwnPermission`
+  - Setting this property to `true` enables the _Credentials/UseOwn_ permission.
+    Primarily useful when using project-based security authorization strategies or authenticated builds, this permission allows an immediate action to access user-scoped credentials from the user's private credentials store.
+    Immediate actions include invoking a build with parameters, tagging a build in SCM, providing credentials to a pipeline input step, and other user interaction.
+    Using this permission can help limit which users are allowed to provide their own credentials as input to an action which can be particularly handy when combined with the Authorize Project plugin where builds may be run as the user who triggered the build.
```
daniel-beck, line 10:
```suggestion
    Using this permission can help limit which users are allowed to provide their own credentials as input to an action which can be particularly handy when combined with the https://plugins.jenkins.io/authorize-project/[Authorize Project] plugin where builds may be run as the user who triggered the build.
```

docs/fflags.adoc
```diff
@@ -0,0 +1,23 @@
...
+    This allows separation of the permission to build a job from the permission to provide user-scoped credentials.
+    When this flag is `false` or unspecified, then _Job/Build_ is used instead for the same purpose.
```
daniel-beck, line 12: As a personal preference, I want to see the default of something early, to better understand the context.

docs/faq.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+=== Where do credentials get stored?
+
+By default, Jenkins uses an internal credentials store where credentials secret data are encrypted at rest in the Jenkins controller file system.
+This security model does not prevent anyone with read access to the controller file system from reading and decrypting those secrets.
+For more sensitive secrets, it is highly recommended to install and use an external credentials provider plugin which delegates the secure storage of secrets to a proper key management system.
+Various cloud providers (including https://plugins.jenkins.io/aws-secrets-manager-credentials-provider/[AWS], https://plugins.jenkins.io/azure-keyvault/[Azure], and https://plugins.jenkins.io/gcp-secrets-manager-credentials-provider/[GCP]) have secrets management and HSMs available for this purpose, and open source software such as https://plugins.jenkins.io/kubernetes-credentials-provider/[Kubernetes], https://plugins.jenkins.io/hashicorp-vault-plugin/[Vault], or https://plugins.jenkins.io/conjur-credentials/[Conjur] provide more generic secrets management for multiple platforms.
```
daniel-beck, line 10: Prefer a list format.

Also, it might be useful to link (with disclaimer about technical content) to the generated list at https://www.jenkins.io/doc/developer/extensions/credentials/#credentialsprovider ?

daniel-beck (commented):
Superficial look so far only, submitting so I don't forget.

docs/faq.adoc
jvz: That link is misleading as some of the external secrets manager plugins use a hybrid approach where secrets are stored in the service, but Jenkins stores credentials internally which are basically placeholders for the external credentials. Due to the various ways that credentials plugins can be made, it seems as though not everyone has used the same approach for this.

docs/fflags.adoc
jvz: I like that preference!

jvz:
@stephenc do you care to take a look through these docs updates?

jeffret-b (approved):
It's great to improve documentation in this area. The other, minor improvements are also good.

I've left some thoughts for consideration about the new documentation. Maybe there are some further possibilities for improvement. I'm also happy to accept these changes as worthwhile improvements on their own.

docs/faq.adoc
```diff
@@ -0,0 +1,74 @@
...
+- https://plugins.jenkins.io/gcp-secrets-manager-credentials-provider/[GCP]
+- https://plugins.jenkins.io/kubernetes-credentials-provider/[Kubernetes]
+- https://plugins.jenkins.io/hashicorp-vault-plugin/[Vault]
+
+A full listing (including some developer-level details) of https://www.jenkins.io/doc/developer/extensions/credentials/[plugins that provide credentials extensions] is available in the developer documentation.
+
+=== How should I organize my credentials?
+
+It's important to understand the access model of credentials in order to organize them most appropriately.
+Credentials are accessed through a combination of scopes and contexts, which are further categorized into types and domains.
```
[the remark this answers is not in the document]
jvz: There's already verbose descriptions of these things in the user guide, but said guide is enormous. Considering barely anyone can get the terminology consistent, I figured explaining it somewhere in the FAQ would help. Perhaps the concepts can be split into "what is _foo_?"

docs/fflags.adoc
```diff
@@ -0,0 +1,23 @@
+== Feature Flags
```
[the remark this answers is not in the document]
jvz: The two permissions aren't going anywhere; they're particularly useful for locking down credentials access with project-based ACLs. The default behavior is to merge these permissions with related but orthogonal ones.

docs/faq.adoc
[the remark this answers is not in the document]
jeffret-b: I'll bet that would work better. I like the idea of getting these basic concepts more accessible to users, but they're still kind of buried inside something else.

PR #167, opened 2020-11-09
[JENKINS-62904] Remove unused files

jeffret-b:
These files are unused and probably broken after the core tables-to-divs migration.

I've searched through the Jenkins ecosystem and haven't found any usages. This looks like something that was started at some point, but never completed, perhaps during the 1.0 to 2.0 transition of credentials-plugin. This doesn't show up in the credentials plugin UI nor is it referenced anywhere else.

After the tables-to-divs migration in Jenkins core, the domainCredential.jelly file probably doesn't work correctly. Instead of removing it, fixing it would require making sure it works correctly before the change and then testing to ensure it has been corrected.

See [JENKINS-62904](https://issues.jenkins.io/browse/JENKINS-62904).


jglick (approved):

jvz (approved):
Same here; I haven't found any usages anymore. This seems to be outdated.

Wadeck (approved):
In the commit [Towards 2.0](https://github.com/jenkinsci/credentials-plugin/commit/8c9e8f8e80348d5592632f4a48ac35f2684403a8), we can see it was used in `SystemCredentialsProvider/index.jelly` but the behavior was changed to have something inline instead of the internal tag. Also it was used [until 2.0.1](https://github.com/jenkinsci/credentials-plugin/commit/db3664cf3317d870d3124b3b9d5ccabbcd3f8625) in `UserCredentialsProvider#UserCredentialsProperty/config.jelly`. 
Finally I checked [the commit](https://github.com/jenkinsci/credentials-plugin/commit/ea696cb21bc84bbe756334ef818450405432410a) introducing the domainCredential.jelly, and confirm the two usages above were the only one introduced at that time.

---

I searched in the ecosystems (OSS + CB): no usage of `domainCredential`.

---

ℹ️ FTR the files in src/main/webapp/help/domain/ are still used, e.g. by CredentialsStoreAction#DomainWrapper.configure.jelly.

PR #205, opened 2021-05-06
[JENKINS-44860] `UsernameCredentials.isUsernameSecret`

jglick:
[JENKINS-44860](https://issues.jenkins.io/browse/JENKINS-44860)

https://github.com/jenkinsci/credentials-binding-plugin/pull/131#issuecomment-832854838

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
   </scm>
 
   <properties>
-    <revision>2.3.19</revision>
+    <revision>2.4</revision>
     <changelist>-SNAPSHOT</changelist>
     <jenkins.version>2.222.4</jenkins.version>
     <java.level>8</java.level>
--- a/pom.xml
+++ b/pom.xml
@@ -70,7 +70,6 @@
     <jenkins.version>2.222.4</jenkins.version>
     <java.level>8</java.level>
     <antlr4.version>4.9.2</antlr4.version>
-    <jcasc.version>1.50</jcasc.version>
   </properties>
 
   <repositories>
@@ -112,7 +111,6 @@
     <dependency>
       <groupId>io.jenkins</groupId>
       <artifactId>configuration-as-code</artifactId>
-      <version>${jcasc.version}</version>
       <optional>true</optional>
     </dependency>
 
@@ -155,7 +153,6 @@
     <dependency>
       <groupId>io.jenkins.configuration-as-code</groupId>
       <artifactId>test-harness</artifactId>
-      <version>${jcasc.version}</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsNameProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsNameProvider.java
@@ -90,7 +90,11 @@ public abstract class CredentialsNameProvider<C extends Credentials> {
         if (nameWith != null) {
             try {
                 CredentialsNameProvider nameProvider = nameWith.value().newInstance();
-                return new Result(nameProvider.getName(credentials), nameWith.priority());
+                String name = nameProvider.getName(credentials);
+                if (!name.isEmpty()) {
+                    LOGGER.fine(() -> "named `" + name + "` from " + nameProvider);
+                    return new Result(name, nameWith.priority());
+                }
             } catch (ClassCastException | InstantiationException | IllegalAccessException e) {
                 // ignore
             }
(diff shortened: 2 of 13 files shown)


src/main/java/com/cloudbees/plugins/credentials/CredentialsNameProvider.java
```diff
@@ -90,7 +90,11 @@ Result name(@NonNull Credentials credentials, @NonNull Class<?> clazz) {
         if (nameWith != null) {
             try {
                 CredentialsNameProvider nameProvider = nameWith.value().newInstance();
-                return new Result(nameProvider.getName(credentials), nameWith.priority());
+                String name = nameProvider.getName(credentials);
+                if (!name.isEmpty()) {
```
jglick, line 94: Make it possible to just `return credentials.getDescription()` and if that happens to be the empty string, still behave reasonably (choose a different provider).

src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
```diff
@@ -53,6 +55,9 @@
     @NonNull
     private final Secret password;
 
+    @Nullable
+    private Boolean usernameSecret = false;
```
jglick, line 59: Default false in the Java constructor which would be used only from tests or scripts.

src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
```diff
@@ -72,6 +77,13 @@ public UsernamePasswordCredentialsImpl(@CheckForNull CredentialsScope scope,
         this.password = Secret.fromString(password);
     }
 
+    private Object readResolve() {
+        if (usernameSecret == null) {
+            usernameSecret = true;
```
jglick, line 82: Default true for deserialized credentials, either from settings predating the upgrade, or REST uploads as XML.

src/main/resources/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl/credentials.jelly
```diff
@@ -28,6 +28,9 @@
   <f:entry title="${%Username}" field="username">
     <f:textbox/>
   </f:entry>
+  <f:entry field="usernameSecret">
+    <f:checkbox title="${%Treat username as secret}"/>
```
jglick, line 32: Default false for new credentials created from the GUI.

Wadeck (approved):
Manually tested:
- ✔️ Existing credentials are not changing behavior
- ✔️ New credentials let the checkbox uncheck (but this could be changed)

Of course, no masking until credentials-masking is merged as well.

Will merge this for next release

bitwiseman:
Why not default to `usernameSecret=true`  in all cases?

jglick:
@bitwiseman https://github.com/jenkinsci/credentials-plugin/pull/205#discussion_r627811232 for the GUI.

car-roll (approved):
I'm fine with making username not secret as default going forward since the behavior for legacy credentials is to keep them hidden.

bitwiseman:
> @bitwiseman https://github.com/jenkinsci/credentials-plugin/pull/205#discussion_r627811232 for the GUI.

I don't understand. Yes, I know it defaults to false in GUI. Why does it default to false? The GUI couldn't default to true/checked?

jglick:
> I know it defaults to false in GUI. Why does it default to false?

Sorry, did not get your question. We expect the username to be nonsecret in most cases, and until now it was displayed in credentials dropdowns to anyone with view configuration permissions. In fact the originally submitted PRs just _unconditionally_ began showing usernames in logs, merely by virtue of updating the `credentials-binding` plugin, which I was uncomfortable with. So here I am trying to walk a compromise path that poses no risk of unwanted surprise for an existing installation, while guiding new configurations to display the username.

pom.xml
```diff
@@ -65,7 +65,7 @@
   </scm>
 
   <properties>
-    <revision>2.4.2</revision>
+    <revision>2.5</revision>
```
jglick, line 68: Just proposing a version bump. Do we really need to keep the micro version? Can revert if desired.

bitwiseman:
> > I know it defaults to false in GUI. Why does it default to false?
> 
> Sorry, did not get your question. We expect the username to be nonsecret in most cases, and until now it was displayed in credentials dropdowns to anyone with view configuration permissions. In fact the originally submitted PRs just _unconditionally_ began showing usernames in logs, merely by virtue of updating the `credentials-binding` plugin, which I was uncomfortable with. So here I am trying to walk a compromise path that poses no risk of unwanted surprise for an existing installation, while guiding new configurations to display the username.

Understood. Sounds good, thanks for the explanation.

bitwiseman (approved):

jglick:
@bitwiseman @car-roll @Wadeck any updates on this & https://github.com/jenkinsci/credentials-binding-plugin/pull/132 & https://github.com/jenkinsci/ssh-credentials-plugin/pull/70?

bitwiseman (approved):

PR #210, opened 2021-05-19
[JENKINS-59109] Bind triggering user's credentials

sbgertp:
Fixes [JENKINS-59109](https://issues.jenkins.io/browse/JENKINS-59109)

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
@@ -971,7 +971,7 @@ public abstract class CredentialsProvider extends Descriptor<CredentialsProvider
      * @return the trigger of the supplied run or {@code null} if this could not be determined.
      */
     @CheckForNull
-    private static Map.Entry<User, Run<?, ?>> triggeredBy(Run<?, ?> run) {
+    public static Map.Entry<User, Run<?, ?>> triggeredBy(Run<?, ?> run) {
         Cause.UserIdCause cause = run.getCause(Cause.UserIdCause.class);
         if (cause != null) {
             User u = User.get(cause.getUserId(), false, Collections.emptyMap());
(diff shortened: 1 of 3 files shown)


jvz (commented):
Not exactly what I had in mind originally, but this might work. I'll take a closer look at this later in the week. In the meantime, maybe @jglick has any feedback?

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
```diff
@@ -971,7 +971,7 @@ public CredentialsStore next() {
      * @return the trigger of the supplied run or {@code null} if this could not be determined.
      */
     @CheckForNull
```
jglick, line 973:
```suggestion
    @Restricted(NoExternalUse.class)
    @CheckForNull
```

We do not expect this to be called from plugins, right? And the use of `Map.Entry` for a tuple is too ugly to bake into any supported API.

jglick (commented):
I do not understand at a conceptual level how user-scoped credentials are supposed to work, so I cannot comment on whether this is “right”. IIUC the effective change here is that `UserIdCause` is being considered not just on the “current” build but also on its transitive upstreams?

src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
jvz: Well, I think the problem here might be that the `build` step will also need to bind a credential parameter binding, so it needs to figure out who triggered that build before propagating it down. Or say we have a pipeline where I bind my user scoped credentials during an `input` step (or as a build parameter) which is needed later on in a `build` step which also takes a build parameter for the credential. The invoked job won't propagate this credential binding info stored in the outer job without implementing the feature I was thinking of for the `build` step. See https://github.com/jenkinsci/credentials-plugin/blob/master/docs/consumer.adoc#binding-user-supplied-credentials-parameters-to-builds for the official API I put together back when I added support to `input` and build parameters for shadowing credentials with user-scoped credentials.

jglick: ^^^ probably meant as a top-level topic, not a comment on this trivial suggestion?

jvz:
Yeah I think this might be backwards. I'd expect a build with a build step to bind its credential parameter bindings to the invoked build.

jvz:
Oh that's interesting! Let me verify this works equivalently to what I had in mind before I can merge this. Should be within the next few days.

PR #225, opened 2021-09-01
[JENKINS-66514] BaseStandardCredentials ID validation error

amuniz:
See details in [JENKINS-66514](https://issues.jenkins.io/browse/JENKINS-66514)

The fix is to remove the unnecessary explicit `checkUrl` so the legacy
validation method isn't used. The check URL will be calculated
automatically and the "new" (years old) mechanism kicks in.

The test is checking that the validation error shows up when needed (and
only when needed), so this change does not cause a regression.


Wadeck (approved):
The removal is fine. The test does not test what was modified in this PR, it's passing even if I revert the removed part.

PR #228, opened 2021-09-21
Code improvements

offa:
- Code clean-up
- Deprecated APIs replaced (still some left)
- Missing nullability annotations added
- JSR 305 annotations replaced

---------------------

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
@@ -388,7 +388,7 @@ public abstract class CredentialsDescriptor extends Descriptor<Credentials> impl
                     if (pathSegments.size() >= 2) {
                         String firstSegment = pathSegments.get(0);
                         if ("user".equals(firstSegment)) {
-                            User user = User.get(pathSegments.get(1));
+                            User user = User.getById(pathSegments.get(1), false);
                             if (type.isInstance(user) && CredentialsProvider.hasStores(user)) {
                                 // we have a winner
                                 return type.cast(user);
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsNameProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsNameProvider.java
@@ -26,6 +26,8 @@ package com.cloudbees.plugins.credentials;
 import com.cloudbees.plugins.credentials.common.IdCredentials;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -89,13 +91,14 @@ public abstract class CredentialsNameProvider<C extends Credentials> {
         NameWith nameWith = clazz.getAnnotation(NameWith.class);
         if (nameWith != null) {
             try {
-                CredentialsNameProvider nameProvider = nameWith.value().newInstance();
+                CredentialsNameProvider nameProvider = nameWith.value().getDeclaredConstructor().newInstance();
                 String name = nameProvider.getName(credentials);
                 if (!name.isEmpty()) {
                     LOGGER.fine(() -> "named `" + name + "` from " + nameProvider);
                     return new Result(name, nameWith.priority());
                 }
-            } catch (ClassCastException | InstantiationException | IllegalAccessException e) {
+            } catch (ClassCastException | InstantiationException | IllegalAccessException
+                    | NoSuchMethodException | InvocationTargetException e) {
                 // ignore
             }
         }
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinition.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinition.java
@@ -4,7 +4,6 @@ import com.cloudbees.plugins.credentials.common.StandardCredentials;
 import com.cloudbees.plugins.credentials.common.StandardListBoxModel;
 import com.cloudbees.plugins.credentials.domains.DomainRequirement;
 import hudson.Extension;
-import hudson.model.Descriptor;
 import hudson.model.Item;
 import hudson.model.ParameterDefinition;
 import hudson.model.ParameterValue;
@@ -125,29 +124,27 @@ public class CredentialsParameterDefinition extends SimpleParameterDefinition {
         public ListBoxModel doFillCredentialTypeItems() {
             ListBoxModel result = new ListBoxModel();
             result.add("Any", StandardCredentials.class.getName());
-            for (Descriptor<Credentials> d : CredentialsProvider.allCredentialsDescriptors()) {
-                if (!(d instanceof CredentialsDescriptor)) {
+            for (CredentialsDescriptor d : CredentialsProvider.allCredentialsDescriptors()) {
+                if (d == null) {
                     continue;
                 }
-                CredentialsDescriptor descriptor = (CredentialsDescriptor) d;
-                if (StandardCredentials.class.isAssignableFrom(descriptor.clazz)) {
-                    result.add(descriptor.getDisplayName(), descriptor.clazz.getName());
+                if (StandardCredentials.class.isAssignableFrom(d.clazz)) {
+                    result.add(d.getDisplayName(), d.clazz.getName());
                 }
             }
             return result;
         }
 
         private Class<? extends StandardCredentials> decodeType(String credentialType) {
-            for (Descriptor<Credentials> d : CredentialsProvider.allCredentialsDescriptors()) {
-                if (!(d instanceof CredentialsDescriptor)) {
+            for (CredentialsDescriptor d : CredentialsProvider.allCredentialsDescriptors()) {
+                if (d == null) {
                     continue;
                 }
-                CredentialsDescriptor descriptor = (CredentialsDescriptor) d;
-                if (!StandardCredentials.class.isAssignableFrom(descriptor.clazz)) {
+                if (!StandardCredentials.class.isAssignableFrom(d.clazz)) {
                     continue;
                 }
-                if (credentialType.equals(descriptor.clazz.getName())) {
-                    return descriptor.clazz.asSubclass(StandardCredentials.class);
+                if (credentialType.equals(d.clazz.getName())) {
+                    return d.clazz.asSubclass(StandardCredentials.class);
                 }
             }
             return StandardCredentials.class;
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinition.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinition.java
@@ -3,6 +3,7 @@ package com.cloudbees.plugins.credentials;
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
 import com.cloudbees.plugins.credentials.common.StandardListBoxModel;
 import com.cloudbees.plugins.credentials.domains.DomainRequirement;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
 import hudson.model.Item;
 import hudson.model.ParameterDefinition;
@@ -116,6 +117,7 @@ public class CredentialsParameterDefinition extends SimpleParameterDefinition {
         /**
          * {@inheritDoc}
          */
+        @NonNull
         @Override
         public String getDisplayName() {
             return Messages.CredentialsParameterDefinition_DisplayName();
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsResolver.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsResolver.java
@@ -25,6 +25,8 @@ package com.cloudbees.plugins.credentials;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
@@ -126,14 +128,14 @@ public abstract class CredentialsResolver<F extends Credentials, T extends Crede
             // if the reflective instantiation proves a hot point, put a cache in front.
             try {
                 @SuppressWarnings("unchecked")
-                final CredentialsResolver<Credentials, C> resolver = resolveWith.value().newInstance();
+                final CredentialsResolver<Credentials, C> resolver = resolveWith.value().getDeclaredConstructor().newInstance();
                 if (Credentials.class.isAssignableFrom(resolver.getFromClass())
                         && clazz.isAssignableFrom(resolver.getToClass())) {
                     return resolver;
                 }
                 LOGGER.log(Level.SEVERE, "Resolver {0} for type {1} resolves to {2} which is not assignable to {1}",
                         new Object[]{resolver.getClass(), clazz, resolver.getToClass()});
-            } catch (InstantiationException | IllegalAccessException e) {
+            } catch (InstantiationException | IllegalAccessException | NoSuchMethodException | InvocationTargetException e) {
                 LOGGER.log(Level.WARNING, "Could not instantiate resolver: " + resolveWith.value(), e);
                 return null;
             }
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
@@ -94,6 +94,7 @@ public class CredentialsSelectHelper extends Descriptor<CredentialsSelectHelper>
     /**
      * {@inheritDoc}
      */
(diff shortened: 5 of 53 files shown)


src/main/java/com/cloudbees/plugins/credentials/CredentialsDescriptor.java
```diff
@@ -388,7 +388,7 @@ public static ModelObject lookupContext(String provider, String token) {
                     if (pathSegments.size() >= 2) {
                         String firstSegment = pathSegments.get(0);
                         if ("user".equals(firstSegment)) {
-                            User user = User.get(pathSegments.get(1));
+                            User user = User.getById(pathSegments.get(1), false);
```
jglick, line 391:
```suggestion
                            User user = User.getById(pathSegments.get(1), true);
```

to preserve non-null semantics of original call

src/main/java/com/cloudbees/plugins/credentials/CredentialsNameProvider.java
```diff
@@ -89,13 +91,14 @@ Result name(@NonNull Credentials credentials, @NonNull Class<?> clazz) {
         NameWith nameWith = clazz.getAnnotation(NameWith.class);
         if (nameWith != null) {
             try {
-                CredentialsNameProvider nameProvider = nameWith.value().newInstance();
+                CredentialsNameProvider nameProvider = nameWith.value().getDeclaredConstructor().newInstance();
```
jglick, line 94:
```suggestion
                CredentialsNameProvider nameProvider = nameWith.value().getConstructor().newInstance();
```

src/main/java/com/cloudbees/plugins/credentials/CredentialsNameProvider.java
```diff
@@ -89,13 +91,14 @@ Result name(@NonNull Credentials credentials, @NonNull Class<?> clazz) {
...
                 String name = nameProvider.getName(credentials);
                 if (!name.isEmpty()) {
                     LOGGER.fine(() -> "named `" + name + "` from " + nameProvider);
                     return new Result(name, nameWith.priority());
                 }
-            } catch (ClassCastException | InstantiationException | IllegalAccessException e) {
+            } catch (ClassCastException | InstantiationException | IllegalAccessException
+                    | NoSuchMethodException | InvocationTargetException e) {
```
jglick, line 101:
```suggestion
            } catch (Exception e) {
```

src/main/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinition.java
```diff
@@ -125,29 +126,27 @@ public String getDisplayName() {
         public ListBoxModel doFillCredentialTypeItems() {
             ListBoxModel result = new ListBoxModel();
             result.add("Any", StandardCredentials.class.getName());
-            for (Descriptor<Credentials> d : CredentialsProvider.allCredentialsDescriptors()) {
-                if (!(d instanceof CredentialsDescriptor)) {
+            for (CredentialsDescriptor d : CredentialsProvider.allCredentialsDescriptors()) {
```
jglick, line 129: No, the original code was correct, it is actually `allCredentialsDescriptors` which is unsafe.

src/main/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinition.java
```diff
@@ -125,29 +126,27 @@ public String getDisplayName() {
...
+                if (d == null) {
```
jglick, line 130: This is not possible.

src/main/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinition.java
```diff
@@ -125,29 +126,27 @@ public String getDisplayName() {
...
                 }
             }
             return result;
         }
 
         private Class<? extends StandardCredentials> decodeType(String credentialType) {
-            for (Descriptor<Credentials> d : CredentialsProvider.allCredentialsDescriptors()) {
-                if (!(d instanceof CredentialsDescriptor)) {
+            for (CredentialsDescriptor d : CredentialsProvider.allCredentialsDescriptors()) {
+                if (d == null) {
```
jglick, line 142: ditto

jglick (commented):
Ran out of time to review further. Will require detailed review from someone experienced in Jenkins code before merging to ensure there are no regressions.

src/main/java/com/cloudbees/plugins/credentials/CredentialsParameterDefinition.java
offa: Reverted (incl. the two comments below).

offa:
Thanks for your review, I have addressed all comments.

src/main/java/com/cloudbees/plugins/credentials/CredentialsNameProvider.java
Wadeck: This was not resolved as the `getDeclaredContructor` is still proposed.

src/main/java/com/cloudbees/plugins/credentials/CredentialsResolver.java
```diff
@@ -126,14 +128,14 @@ protected CredentialsResolver(@NonNull Class<F> fromClass) {
             // if the reflective instantiation proves a hot point, put a cache in front.
             try {
                 @SuppressWarnings("unchecked")
-                final CredentialsResolver<Credentials, C> resolver = resolveWith.value().newInstance();
+                final CredentialsResolver<Credentials, C> resolver = resolveWith.value().getDeclaredConstructor().newInstance();
```
Wadeck, line 131: ⁉️ Same here, not sure to understand the added value. More verbose + more exceptions.

src/main/java/com/cloudbees/plugins/credentials/CredentialsNameProvider.java
offa: Sorry, seems I've missed that.

> ⁉️ Not sure to understand the improvement here.

`newInstance()` is deprecated and `getDeclaredConstructor().newInstance()` is suggested by the docs.

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
```diff
@@ -577,8 +578,8 @@ public WrappedCredentialsStore getContext(String token) {
          * @param token    the context's {@link ContextResolver#getToken(ModelObject)}.
          * @param store    the {@link CredentialsStore}
          */
-        public WrappedCredentialsStore(ContextResolver resolver, CredentialsProvider provider,
-                                       String token, CredentialsStore store) {
+        public WrappedCredentialsStore(@NonNull ContextResolver resolver, @NonNull CredentialsProvider provider,
+                                       @NonNull String token, CredentialsStore store) {
```
Wadeck, line 582: ⁉️ From the usage inside `WrappedContextResolverCredentialsProvider#getContext` I am OK with resolver and provider being nonNull, but token is not obvious (but seems to be) and store is also nonNull. Was this proposed by IntelliJ or a manual analysis? (not proposed by my IntelliJ configuration ^^)

So, why `store` is not marked in this case?

src/main/java/com/cloudbees/plugins/credentials/CredentialsResolver.java
offa, line 131: Should this become a `getConstructor()` too?

src/main/java/com/cloudbees/plugins/credentials/UserCredentialsProvider.java
```diff
@@ -679,12 +678,13 @@ public boolean hasPermission(@NonNull Authentication a, @NonNull Permission perm
...
          */
+        @NonNull
         @Override
         public ACL getACL() {
             return new ACL() {
                 @Override
-                public boolean hasPermission(@Nonnull Authentication a, @Nonnull Permission permission) {
-                    return user.equals(User.get(a.getName())) && user.getACL().hasPermission(a, permission);
+                public boolean hasPermission(@NonNull Authentication a, @NonNull Permission permission) {
+                    return user.equals(User.getById(a.getName(), false)) && user.getACL().hasPermission(a, permission);
```
Wadeck, line 687:
```suggestion
                    return user.equals(User.getById(a.getName(), true)) && user.getACL().hasPermission(a, permission);
```

❌ Like the one proposed by Jesse above.

src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
```diff
@@ -102,7 +102,7 @@ public String getUsername() {
 
     @Override
     public boolean isUsernameSecret() {
-        return usernameSecret;
+        return Boolean.TRUE.equals(usernameSecret);
```
Wadeck, line 105: 💡 FYI `Unboxing of 'usernameSecret' may produce 'NullPointerException'` is a false positive here as it's not possible to have that value being null, due to the `readResolve` method.

But still OK to have this change.

Wadeck (commented):
👍 For Lambda, NonNull refactoring, import optimization it seems to be fine.  (those are proposed by IntelliJ)

⁉️ But for `getDeclaredConstructor`, I will not merge this without an explanation of the benefit using this approach, it seems less interesting than current version. Perhaps it's just me not knowing why you are doing that ;-)
So, either you provide the advantage/rationale of that approach or you remove them from this PR so that I can merge it.

💡 In general, I would say that I prefer to have PRs without a mix of manual proposals and IntelliJ (or other tool) proposals. The manual ones require more review than the automated ones. And saying it came from a tool in the PR description also helps ;)

Thanks for your contribution @offa 👍

src/main/java/com/cloudbees/plugins/credentials/CredentialsResolver.java
Wadeck: My advice is to just revert the modifications related to newInstance/constructors and I will be able to merge your PR :)

Wadeck:
If you want me to update your PR and merge it just tell me ;) 
(without constructor changes, with the NonNull annotation on WrappedCredentialsStore changed and the getById changed)

src/main/java/com/cloudbees/plugins/credentials/CredentialsResolver.java
offa: I'm fine with that too 👍; fixup commit is on the way ...

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
offa: The change reflects the annotations of the fields to the constructor, as it's more visible to the caller. While `token` (line 567) is marked with `@NonNull`, `store` (line 571) isn't. From it's usage it assumes non-null, but I wasn't confident enough to actually mark it.

offa:
> ⁉️ But for getDeclaredConstructor, I will not merge this without an explanation of the benefit using this approach, it seems less interesting than current version. Perhaps it's just me not knowing why you are doing that ;-)
So, either you provide the advantage/rationale of that approach or you remove them from this PR so that I can merge it.

I'm on Java 11+ and `newInstance()` is deprecated since Java 9 ([Java 11 docs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#newInstance())). The docs (and [discussion on SO](https://stackoverflow.com/questions/46393863/what-to-use-instead-of-class-newinstance)) suggest `getDeclaredConstructor().newInstance()` as the replacement.

However, it's better to get a consensus here, so I'll remove it from this PR and in doubt provide a follow up.

offa:
> If you want me to update your PR and merge it just tell me ;)
(without constructor changes, with the NonNull annotation on WrappedCredentialsStore changed and the getById changed)

Thanks 😄, but I've committed `getById()` through PR suggestion, `newInstance()` is reverted (I'll push once tested locally). Regarding annotation, please see my comment above.

src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
offa: I see. However, the field is annotated as nullable and there's setter accepting `null`, so in theory a NPE is possible.

src/main/java/com/cloudbees/plugins/credentials/CredentialsResolver.java
offa: Resolving here, change reverted to `newInstance()`.

Wadeck:
> However, it's better to get a consensus here, so I'll remove it from this PR and in doubt provide a follow up.

Perfect yes thank you :) 

> regarding annotation, please see my comment above.

Will check the latest version and adjust if necessary before merge 👍 

> I'm on Java 11+ [...]

Thanks for the link and the rationale, will keep that in a place of my brain for next reviews / plugins 👍

src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
Wadeck: If you mean: https://github.com/jenkinsci/credentials-plugin/blob/a2b5c5fba5b7fe77854efeab539b5e4843d52719/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java#L108-L111

It does not accept `null` as it's the boolean primitive.

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
Wadeck: I will adjust the field at the same time, I thought you were changing the constructor by looking at the only location it was called ;)

Wadeck:
Will merge once the CI is good :)

Wadeck:
@offa Thanks for the contribution! Don't hesitate to make others, Hacktoberfest started today, you can gain a t-shirt with 5 approved PRs during October, more info on https://hacktoberfest.digitalocean.com/

jglick:
Use `getConstructor()`, not `getDeclaredConstructor()`, since in the contexts in which we would be calling `newInstance()` we expect the no-arg constructor to be `public`.

src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
offa: :man_facepalming: …

jglick: Revert this misleading change and use `@SuppressFBWarnings` if necessary, though IIUC it should _not_ be necessary since I marked the field `@Nullable` not `@CheckForNull` to begin with.

src/test/java/com/cloudbees/plugins/credentials/CredentialsProviderTest.java
```diff
@@ -278,7 +276,7 @@ public void testNodeCredentialFingerprintsAreRemovedForNonExistentNodes() throws
         // it should not be recorded
         DumbSlave nonAddedSlave = new DumbSlave("non-added-slave",
                 "dummy", "/home/test/agent", "1", Node.Mode.NORMAL, "remote",
-                new JNLPLauncher(),
+                new JNLPLauncher(false),
```
jglick, line 279: This was a bad change in core BTW. The no-arg ctor should have been retained, and `enableWorkDir` made into a `@DataBoundSetter`, or just removed altogether since it is a derivative concept. Way on the back burner to fix, but I would suggest reverting this and just suppressing the deprecation warning.

jglick:
(still reviewing)

src/test/java/com/cloudbees/plugins/credentials/CredentialsUnavailableExceptionTest.java
```diff
@@ -205,10 +202,9 @@ public boolean requiresWorkspaceForPolling() {
         }
 
         @Override
-        public void checkout(@Nonnull Run<?, ?> build, @Nonnull Launcher launcher, @Nonnull FilePath workspace,
-                             @Nonnull TaskListener listener, @javax.annotation.CheckForNull File changelogFile,
-                             @javax.annotation.CheckForNull SCMRevisionState baseline)
-                throws IOException, InterruptedException {
+        public void checkout(@NonNull Run<?, ?> build, @NonNull Launcher launcher, @NonNull FilePath workspace,
+                             @NonNull TaskListener listener, @javax.annotation.CheckForNull File changelogFile,
+                             @javax.annotation.CheckForNull SCMRevisionState baseline) {
```
jglick, line 207:
```suggestion
                             @CheckForNull SCMRevisionState baseline) {
```

?

jglick (commented):
Looks good other than a couple of nits. Thanks!

src/test/java/com/cloudbees/plugins/credentials/CredentialsProviderTest.java
offa: Both done.

src/test/java/com/cloudbees/plugins/credentials/CredentialsUnavailableExceptionTest.java
offa: Nice catch!

jglick (approved):

PR #230, opened 2021-09-22
feat: [JENKINS-66699] Exchange default PNGs with SVG icons

NotMyFault:
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant issues in GitHub or Jira
- [X] Link to relevant pull requests, esp. upstream and downstream changes
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue

This PR aims to replace the PNGs with SVGs to present a more modern looking-like appearance.
Addresses https://issues.jenkins.io/browse/JENKINS-66699

**Before**:

**After**:


src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/CredentialsWrapper/sidepanel.jelly
```diff
@@ -27,10 +27,10 @@
   <l:side-panel>
     <l:tasks>
     <j:set var="url" value="${h.getNearestAncestorUrl(request,it)}"/>
-      <l:task icon="images/24x24/up.png" href="../.." title="${%backto(it.domain.displayName)}" contextMenu="false"/>
-      <l:task icon="images/24x24/setting.png" href="${url}/update" title="${%Update}" permission="${it.domain.parent.UPDATE}"/>
-      <l:task icon="images/24x24/edit-delete.png" href="${url}/delete" title="${%Delete}" permission="${it.domain.parent.DELETE}"/>
```
jglick, line 32: These are coming from core so they should remain as is, at least until the core dep is new enough that we can use SVGs from https://github.com/jenkinsci/jenkins/pull/5663.

src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/DomainWrapper/sidepanel.jelly
```diff
@@ -27,11 +27,11 @@
   <l:side-panel>
     <l:tasks>
     <j:set var="url" value="${h.getNearestAncestorUrl(request,it)}"/>
-    <l:task icon="images/24x24/up.png" href="../.." title="${%Back to credential domains}" contextMenu="false"/>
-      <l:task icon="plugin/credentials/images/24x24/new-credential.png" href="newCredentials" title="${%Add Credentials}" permission="${it.parent.CREATE}" />
+    <l:task icon="plugin/credentials/images/up.svg" href="../.." title="${%Back to credential domains}" contextMenu="false"/>
+      <l:task icon="plugin/credentials/images/new-credential.svg" href="newCredentials" title="${%Add Credentials}" permission="${it.parent.CREATE}" />
       <j:if test="${!it.global and it.parent.domainsModifiable}">
-      <l:task icon="images/24x24/setting.png" href="${url}/configure" title="${%Configure}" permission="${it.parent.MANAGE_DOMAINS}"/>
-      <l:task icon="images/24x24/edit-delete.png" href="${url}/delete" title="${%Delete domain}" permission="${it.parent.MANAGE_DOMAINS}"/>
```
jglick, line 34: ditto

jglick (commented):
SVG renames seem good. AFAICT there should be no _new_ SVGs in this PR, because we should be picking them up from core sooner or later.

I suppose `src/main/webapp/images/make-inkscape.sh` should not get bundled in the binary.

@timja @uhafner @oleg-nenashev as reviewers of https://github.com/jenkinsci/jenkins/pull/5663

src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/CredentialsWrapper/sidepanel.jelly
timja: Few options here I guess:

1. bump the minimum now
2. add a method to CredentialsWrapper that determines if the version is new enough and then choose dynamically
3. do this, which temporarily copies the svgs to this plugin
4. leave with non sharp pngs, not ideal

jglick: > bump the minimum now

To at least 2.308 I guess. I am inclined to do that as it is the simplest option.

NotMyFault: When I created this PR, I've chosen to use approach 3, hence there are additions. I thought that would be a better temporary and lesser evil solution, until LTS includes 2.308 and you can rely on SVGs natively.

jglick: > until LTS includes 2.308

I do not think we care. It is fine to release the plugin with a dep on a weekly. After all, users who really care about PNG vs. SVG would be unhappy with the current LTS anyway!

NotMyFault: True I guess, if you like blurry images, you should stay on the LTS (for now) with the PNGs.

timja:
> I suppose src/main/webapp/images/make-inkscape.sh should not get bundled in the binary.

I think we removed the script from jenkinsci/jenkins as part of the original PR, it can probably be removed from here too?

NotMyFault:
> > I suppose src/main/webapp/images/make-inkscape.sh should not get bundled in the binary.
> 
> I think we removed the script from jenkinsci/jenkins as part of the original PR, it can probably be removed from here too?

I've removed it for now and removed the added svgs in favor of increasing the minimum version to 2.308.

src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/CredentialsWrapper/sidepanel.jelly
```diff
@@ -27,10 +27,10 @@
...
-      <l:task icon="plugin/credentials/images/24x24/move.png" href="${url}/move" title="${%Move}" permission="${it.domain.parent.DELETE}"/>
+      <l:task icon="images/svgs/up.svg" href="../.." title="${%backto(it.domain.displayName)}" contextMenu="false"/>
```
jglick, line 30: So IIUC this will not work with the current `jenkins.version` in `pom.xml`?

timja:
You need to update `jenkins.version` in pom.xml for that

src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/CredentialsWrapper/sidepanel.jelly
NotMyFault: The SVG icons are only available on versions equal or higher than 2.308. I just pushed a change that addresses that.

jglick (approved):
Diff looks good. Needs an outside sanity check. Not sure who the plugin maintainer here is.

timja:
generally @jvz, but also @daniel-beck and @Wadeck by the looks of it

oleg-nenashev:
I'm not sure I am still a maintainer. IIRC @daniel-beck has cleaned up my name at some point. Otherwise can release it

timja:
> Looks good to me. Will need an upgrade guide entry in the Jenkins core

Probably overkill? If users don't do this they will see what they had before. When they upgrade they will get sharper icons

Wadeck (approved):
Except the user credentials icon, the SVGs are better, especially at higher level of zoom.

Thanks for the PR @NotMyFault 

<details>

<summary>Screenshots</summary>

### 100%, before

before_100

### 100%, after

after_100


### 200%, before


before_200

### 200%, after


after_200

</details>

daniel-beck:
Likely caused [JENKINS-68674](https://issues.jenkins.io/browse/JENKINS-68674).

PR #235, opened 2021-10-01
Add config as code support for credentials configuration

timja:
Pending incremental from jcasc for the `export` test to pass
Importing works fine with this PR alone

some minor jcasc testing cleanup done as well

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
   <properties>
     <revision>2.6.2</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.222.4</jenkins.version>
+    <jenkins.version>2.249.1</jenkins.version>
     <java.level>8</java.level>
     <antlr4.version>4.9.2</antlr4.version>
   </properties>
@@ -89,8 +89,8 @@
     <dependencies>
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
-        <artifactId>bom-2.222.x</artifactId>
-        <version>887.vae9c8ac09ff7</version>
+        <artifactId>bom-2.249.x</artifactId>
+        <version>950.v396cb834de1e</version>
         <scope>import</scope>
         <type>pom</type>
       </dependency>
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsTypeFilter.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsTypeFilter.java
@@ -35,7 +35,6 @@ import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.DataBoundConstructor;
@@ -199,11 +198,7 @@ public abstract class CredentialsTypeFilter extends AbstractDescribableImpl<Cred
          */
         @NonNull
         public List<String> getClassNames() {
-            return ExtensionList.lookup(CredentialsDescriptor.class)
-                    .stream()
-                    .map(type -> type.getClass().getName())
-                    .filter(classNames::contains)
-                    .collect(Collectors.toList());
+            return new ArrayList<>(classNames);
         }
 
         /**
@@ -311,11 +306,7 @@ public abstract class CredentialsTypeFilter extends AbstractDescribableImpl<Cred
          */
         @NonNull
         public List<String> getClassNames() {
-            return ExtensionList.lookup(CredentialsDescriptor.class)
-                    .stream()
-                    .map(type -> type.getClass().getName())
-                    .filter(classNames::contains)
-                    .collect(Collectors.toList());
+            return new ArrayList<>(classNames);
         }
 
         /**
--- a/src/test/resources/com/cloudbees/plugins/credentials/casc/credentialsProviderManager.yaml
+++ b/src/test/resources/com/cloudbees/plugins/credentials/casc/credentialsProviderManager.yaml
@@ -1,0 +1,14 @@
+globalCredentialsConfiguration:
+  configuration:
+    providerFilter:
+      excludes:
+        classNames:
+          - "com.cloudbees.plugins.credentials.UserCredentialsProvider"
+    restrictions:
+      - includes:
+          provider: "com.cloudbees.hudson.plugins.folder.properties.FolderCredentialsProvider"
+          type: "com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl$DescriptorImpl"
+    typeFilter:
+      excludes:
+        classNames:
+          - "com.cloudbees.plugins.credentials.impl.CertificateCredentialsImpl$DescriptorImpl"
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsProviderFilter.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsProviderFilter.java
@@ -31,6 +31,7 @@ import hudson.ExtensionPoint;
 import hudson.Util;
 import hudson.model.AbstractDescribableImpl;
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
@@ -159,7 +160,7 @@ public abstract class CredentialsProviderFilter extends AbstractDescribableImpl<
     }
 
     /**
-     * A filter that implements a allowlist policy, "if you are not on the list you can't come in".
+     * A filter that implements an allow-list policy, "if you are not on the list you can't come in".
      *
      * @since 2.0
      */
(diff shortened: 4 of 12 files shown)


pom.xml
```diff
@@ -67,7 +67,7 @@
   <properties>
     <revision>2.6.2</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.222.4</jenkins.version>
+    <jenkins.version>2.249.1</jenkins.version>
```
timja, line 70: jcasc baseline

src/main/java/com/cloudbees/plugins/credentials/CredentialsTypeFilter.java
```diff
@@ -199,11 +198,7 @@ public boolean filter(CredentialsDescriptor type) {
          */
         @NonNull
         public List<String> getClassNames() {
-            return ExtensionList.lookup(CredentialsDescriptor.class)
-                    .stream()
-                    .map(type -> type.getClass().getName())
-                    .filter(classNames::contains)
-                    .collect(Collectors.toList());
+            return new ArrayList<>(classNames);
```
timja, line 201: I couldn't find any references to this method anywhere so I don't know how to test if this works or not, all tests in this plugin pass.

timja: I've checked and the functionality for includes and excludes still works:

image

pom.xml
Wadeck: Due to #230, this will be 2.308 ;)

timja: aware but depends which is merged first, trivial conflict resolution

timja: ah i see merged

src/main/java/com/cloudbees/plugins/credentials/CredentialsTypeFilter.java
Wadeck: 💡 What you tested in your screenshot is the `CredentialsProviderFilter`. 
For `CredentialsTypeFilter` it's the Types > Add > Includes / Excludes.
Manually tested, it still works 👍 (at least, not discovered flaw) 

(the last one, Restrictions, is covered by CredentialsProviderTypeRestriction)

Wadeck (approved):
I confirm all tests are passing except the export.

Waiting on the companion PR to merge, but I will try to merge other PRs and releasing before to not have incrementals in release.

jetersen (approved):
Overall looks good to me :)

timja:
> Waiting on the companion PR to merge, but I will try to merge other PRs and releasing before to not have incrementals in release.

Propogating through atm, will merge and release bom and update bom, if you're in no rush I'd hold off

pom.xml
```diff
@@ -89,8 +89,8 @@
     <dependencies>
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
-        <artifactId>bom-2.222.x</artifactId>
-        <version>887.vae9c8ac09ff7</version>
+        <artifactId>bom-2.303.x</artifactId>
+        <version>950.v396cb834de1e</version>
```
timja, line 93:
```suggestion
        <version>966.v3857b7c82032</version>
```

Wadeck:
> if you're in no rush I'd hold off

Doing some reviews/merge today, if not done today, will come back on it soon-ish after ;)

src/test/resources/com/cloudbees/plugins/credentials/casc/credentialsProviderManager.yaml
```diff
@@ -0,0 +1,14 @@
+globalCredentialsConfiguration:
+  configuration:
+    providerFilter:
+      excludes:
+        classNames:
+          - "com.cloudbees.plugins.credentials.UserCredentialsProvider"
+    restrictions:
+      - includes:
+          provider: "com.cloudbees.hudson.plugins.folder.properties.FolderCredentialsProvider"
+          type: "com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl$DescriptorImpl"
```
jglick, line 10: Hmm, nothing in JCasC should ever be referring to a `Descriptor` class name. It is not even required that you implement `Descriptor` with a distinct type—you can use a static field or an anonymous inner class or whatever. While we may be required to allow such an entry in a historical `config.xml`, going forward, we should be using `Descriptor.getId()`, which is _usually_ (but not necessarily) a `Describable` class name.

timja: any psuedo-code adapting hint? or an example?

jglick: Do you mean for the `readResolve` impl, or for using `Descriptor.id` going forward?

timja: descriptor.id

timja: do you mean just store the descriptor id instead of the full class name?

jglick: Right, so in this example `com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl` rather than `com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl$DescriptorImpl`.

timja: or the symbol if it has one? `usernamePassword` ? I assume can be looked up pretty easily

timja: The UI is built off of https://github.com/jenkinsci/credentials-plugin/blob/master/src/main/resources/com/cloudbees/plugins/credentials/CredentialsProviderManager/Configuration/config.jelly#L30

timja: not sure much how it works need to track it back

jglick: > or the symbol if it has one?

If you like.

timja: Hmm I'm not sure if this functionality is broken or just only affects the UI.

I've made the changes locally (although no backwards compat added yet).

From my testing the only thing that was affected was the doFill credentials from credentials binding. (probably anything using the credentials jelly control) Running a job where you'd already selected the credential and then restricted it works just fine. (Or using pipeline and free text filling it in)

Anyone know if that's intended?

Docs appear to be here: https://github.com/jenkinsci/credentials-plugin/blob/master/docs/user.adoc
Although there's no inline help

jvz: Yes that's intended. The include and exclude filters are there for making the UI easier to use. Permissions are what control usage restrictions.

timja: Include and exclude work how I expect and prevent them being used, restrictions is the weird one but ok if it’s UI only…

src/test/java/com/cloudbees/plugins/credentials/CredentialsProviderTypeRestrictionTest.java
```diff
@@ -0,0 +1,43 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+
+public class CredentialsProviderTypeRestrictionTest {
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    @Test
+    @LocalData
```
timja, line 22: I don't understand what's going on here,

During the test:

image

At runtime:

image

Last night when I was playing with it at runtime it was working fine, the descriptor was loaded properly and the adapting code was working just fine =/

timja: After wiping my work directory it's back to working at runtime:

image

src/test/resources/com/cloudbees/plugins/credentials/casc/credentialsProviderManager.yaml
timja: Something weird going on with descriptor's not being loaded =/, https://github.com/jenkinsci/credentials-plugin/pull/235/files#r721098631

Any ideas?
Otherwise maybe drop the test? Or even the `readResolve` given this seems to be a UI only thing.

src/test/java/com/cloudbees/plugins/credentials/CredentialsProviderTypeRestrictionTest.java
timja: Actually I'll just use the dummy credentials as that seems to be what other tests do, still weird it doesn't work

src/test/resources/com/cloudbees/plugins/credentials/casc/credentialsProviderManager.yaml
timja: Actually I'll just use the dummy credentials as that seems to be what other tests do, still weird it doesn't work

jglick (approved):

jetersen (approved):

jglick:
Possibly caused: https://issues.jenkins.io/browse/JENKINS-67170

PR #238, opened 2021-10-11
Enable JDK11 in Jenkinsfile

aneveux:
Hello :wave: 

Here is a simple PR enabling JDK11 in the Jenkinsfile of this plugin, helping the global support of Java 11 in Jenkins plugins.

Please let me know if anything needs to be updated or if I missed something. Thanks a lot!

On my side, I validated that compilation + tests were running fine with JDK11. I could not validate the execution on Jenkins though, hence just modifying the Jenkins file and following the CI execution here.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,5 @@
-buildPlugin()
+buildPlugin(configurations: [
+                [platform: 'linux', jdk: '8'],
+                [platform: 'linux', jdk: '11'],
+                [platform: 'windows', jdk: '11'],
+            ])


jglick:
I do not think we should waste time running tests using two different Java versions on Windows. We have enough problems keeping the Windows agents running as it is, without adding extra load. So long as at least one branch tests Linux, at least one Windows, at least one JDK 8, and at least one JDK 11, all should be fine; it is extremely unlikely that there would be undetected bug affecting only Windows _and_ only on a specific Java version. Actually I am inclined to think even the current matrix is wasting CI resources—my preference would be for Linux/11 + Windows/8, to minimize resource usage while retaining good coverage. (Better than Linux/8 + Windows/11 because you do not want all the weirdness in one branch—if you see a strange test failure that might be platform-specific, you want to be able to tell immediately if the problem is Windows or the problem is Java 11.)

PR #248, opened 2021-11-12
Refer to icons using classes instead of filenames

zbynek:
SVG icons from core (added in https://github.com/jenkinsci/credentials-plugin/pull/230 ) can be referenced by class instead of depending on file location in core. That should be more future-proof and seems to be the preferred way.

- [n/a] Link to relevant issues in GitHub or Jira
- [n/a] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
   <properties>
     <revision>2.6.3</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.289.1</jenkins.version>
+    <jenkins.version>2.303.1</jenkins.version>
     <java.level>8</java.level>
     <antlr4.version>4.9.2</antlr4.version>
   </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
   <properties>
     <revision>2.6.3</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.308</jenkins.version>
+    <jenkins.version>2.289.1</jenkins.version>
     <java.level>8</java.level>
     <antlr4.version>4.9.2</antlr4.version>
   </properties>
--- a/src/main/webapp/images/credential.svg
+++ b/src/main/webapp/images/credential.svg
@@ -12,7 +12,6 @@
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
    inkscape:export-ydpi="90"
    inkscape:export-xdpi="90"
-   inkscape:export-filename="/Users/stephenc/src/plugins/cloudbees-registration-plugin/src/main/webapp/images/48x48/credentials.png"
    sodipodi:docname="credentials.svg"
    inkscape:version="0.48.2 r9819"
    sodipodi:version="0.32"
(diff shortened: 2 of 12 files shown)


zbynek:
CC @NotMyFault @Wadeck @jglick

NotMyFault (approved):
Looks good to me, but it works as well on a lower baseline such as 2.277, based on my testing.

src/main/webapp/images/credential.svg
```diff
@@ -12,7 +12,6 @@
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
    inkscape:export-ydpi="90"
    inkscape:export-xdpi="90"
-   inkscape:export-filename="/Users/stephenc/src/plugins/cloudbees-registration-plugin/src/main/webapp/images/48x48/credentials.png"
```
jglick, line 15: Fine, but why here?

pom.xml
```diff
@@ -67,7 +67,7 @@
   <properties>
     <revision>2.6.3</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.308</jenkins.version>
+    <jenkins.version>2.289.1</jenkins.version>
```
jglick, line 70: Does not match https://github.com/jenkinsci/credentials-plugin/blob/15eb87b73807f41a5c6fadf59521ba755262d538/pom.xml#L92

jglick (changes requested):
Other than the BOM mistake, seems OK in principle, assuming it has been tested and indeed works as expected both on the newly lower baseline and on current weeklies. The icon class change seems right; the baseline version change seems unnecessary, especially given that we are already accommodating current plugin releases in https://github.com/jenkinsci/bom/pull/691.

pom.xml
zbynek: Thanks, bumped to 2.303.1

pom.xml
```diff
@@ -67,7 +67,7 @@
   <properties>
     <revision>2.6.3</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.308</jenkins.version>
+    <jenkins.version>2.303.1</jenkins.version>
```
jglick, line 70: Still believe this is unnecessary (generally prefer to build & test against the version which actually forced the original change) but :shrug: either should be OK.

jglick (approved):
Looks OK. Not tested.

pom.xml
NotMyFault: At this point we can even use the next LTS baseline, but eh.

jglick: So
```suggestion
    <jenkins.version>2.308</jenkins.version>
```
and just reword PR to say that it is improving icon display, or whatever it is. Not really up to speed on icon handling in Jenkins so I am not sure what the practical impact is of this change. Do not care about publishing updates for 2.303.x users; if they want updates for anything, start by moving to 2.319.x.

zbynek: Thanks for the feedback. Core version reverted and decription updated.

jglick:
CI problem with Artifactory, trying to trigger a fresh build (for some reason the gesture on the **Checks** tab is not working)…

timja:
> CI problem with Artifactory, trying to trigger a fresh build (for some reason the gesture on the **Checks** tab is not working)…

If checkout fails then blue ocean replay and checks re-run doesn't work. Needs a build now clicked or this

PR #249, opened 2021-11-13
[JENKINS-67132] Fix NPE in isUsernameSecret

MarkEWaite:
[JENKINS-67132](https://issues.jenkins.io/browse/JENKINS-67132) reports a  null pointer exception reading the nullable field isUsernameSecret.  Check the field for null before returning it as a boolean type.

Did not provide a test.  If one is required, will work on that later

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
@@ -104,7 +104,7 @@ public class UsernamePasswordCredentialsImpl extends BaseStandardCredentials imp
 
     @Override
     public boolean isUsernameSecret() {
-        return usernameSecret;
+        return usernameSecret == null ? false : usernameSecret;
     }
 
     @DataBoundSetter


jglick (commented):
How is this reproduced? https://github.com/jenkinsci/credentials-plugin/blob/f09ec61224ef98165ee942349b5c20d670a7820b/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java#L61 and https://github.com/jenkinsci/credentials-plugin/blob/f09ec61224ef98165ee942349b5c20d670a7820b/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java#L82-L87 ought to have covered all the possibilities.

daniel-beck:
Per my comment in https://issues.jenkins.io/browse/JENKINS-67132, visibility of https://github.com/jenkinsci/credentials-plugin/blob/f09ec61224ef98165ee942349b5c20d670a7820b/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java#L82 might be the cause for https://github.com/search?q=org%3Ajenkinsci+%22extends+UsernamePasswordCredentialsImpl%22&type=code

The report mentions Openshift/Openstack, which could be at a glance `OpenShiftBearerTokenCredentialImpl` and `OpenstackKeystoneV3`.

src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
```diff
@@ -104,7 +104,7 @@ public String getUsername() {
 
     @Override
     public boolean isUsernameSecret() {
-        return usernameSecret;
+        return usernameSecret == null ? false : usernameSecret;
```
jglick, line 107: Wrong default.

jglick (changes requested):
I am preparing a proper fix, with test.

MarkEWaite:
Closing this pull request since @jglick will provide a better fix.  Thanks @jglick !

PR #255, opened 2021-12-12
Improve UI with modernise form in core

timja:
see https://github.com/jenkinsci/jenkins/pull/5923

I don't think this will be compatible with the UI so will need to bump baseline once released

image

image


jglick:
Any relation to #287?

timja:
> Any relation to #287?

Yeah this was an earlier fix although pretty sure I fixed it in #287 as well having completely forgotten about this PR.

PR #287, opened 2022-03-05
Modernise UI

timja:
Requires https://github.com/jenkinsci/jenkins/pull/6307

Principles applied:
- All pages have a heading
- Title matches the heading
- All buttons say what they do instead of OK
- All buttons on a new line not next to text
- All primary buttons are styled appropriately
- Switch to core controls where possible otherwise backport changes

TODO:
- Get upstream PR merged

image

<details>
<summary>Manage Jenkins</summary>

image

</details>

<details>
<summary>Add credential button spacing fixed</summary>

image

</details>

<details>
<summary>Add credential dialog</summary>

image
image

</details>

<details>
<summary>Global credentials list</summary>

image

</details>

<details>
<summary>New Credentials</summary>

image

</details>

<details>
<summary>Update credentials</summary>

image

</details>

<details>
<summary>Move credentials</summary>

image

</details>

<details>
<summary>Delete credentials</summary>

image

</details>

<details>
<summary>Add Domain</summary>

image

</details>

<details>
<summary>New domain</summary>

image

</details>

<details>
<summary>View domain</summary>

image

</details>

<details>
<summary>Update domain</summary>

image

</details>

<details>
<summary>Delete domain</summary>

image

</details>

<details>
<summary>Credentials parameter</summary>

image

</details>

PR #291, opened 2022-03-23
Update Jenkins baseline to 2.319.x

jglick:
Cleanup from #230.


timja (approved):
is fine but see also https://github.com/jenkinsci/credentials-plugin/pull/287 which bumps to 2.340

PR #293, opened 2022-03-23
Introduce `Credentials.forRun` to contextualize secrets

jglick:
https://github.com/jenkinsci/github-branch-source-plugin/pull/527 is a demonstrated use case, and https://github.com/jenkinsci/conjur-credentials-plugin/issues/21 a proposed one.

This new SPI requires use of the `CredentialsProvider.findCredentialById` method, which passes a (mandatory) `Run` argument and is the normal way to look up a particular credential during a build. (`credentials-binding` has long used it—https://github.com/jenkinsci/credentials-binding-plugin/pull/169 is the integration test—and the revised version of https://github.com/jenkinsci/git-plugin/pull/1242 is able to use it consistently as well.) There are other lookup methods which take an `Item` (~ `Job`) or even `ItemGroup` (~ `Jenkins` / `Folder` / `MultiBranchProject` / `OrganizationFolder`) context, which would suffice for https://github.com/jenkinsci/github-branch-source-plugin/pull/527 (which would usually be inspecting a `MultiBranchProject` though I left in support for standalone `Job` projects using the `github` plugin’s project property as well). If and when such an SPI becomes desirable, it should be possible to compatibly introduce `Credentials.forItem` (defaulting to `this`, rewrite the default of `forRun` to delegate to `forItem(run.getParent())`) or even `forItemGroup` (similar delegation chain).

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/Credentials.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/Credentials.java
@@ -27,6 +27,7 @@ import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.ExtensionPoint;
 import hudson.model.Describable;
+import hudson.model.Run;
 import java.io.Serializable;
 
 /**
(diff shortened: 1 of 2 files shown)


timja (approved):

MarkEWaite (approved):
I've run it in my Jenkins test instance and have not detected any issues.  I have not exercised contextualized secrets yet in that instance.

PR #295, opened 2022-03-25
Add CredentialsUseListener to improve tracking of Credentials usage

meiswjn:
This change will allow tracking Credentials usage much easier, independent from Fingerprints. So far it was necessary to use the SaveableListener provided by the Fingerprints to subscribe to Credentials usage. Because certain credential consumers do not create fingerprints, this approach is flawed.

Creating the listener gives a lightweight method of tracking Credentials usage, improving overall security.

Tests are not provided, as the logic of the abstract functions is to be implemented by plugins using this approach. Tests are included in the plugin mentioned below.

Usecase: https://github.com/jenkinsci/audit-trail-plugin/pull/68


src/main/java/com/cloudbees/plugins/credentials/CredentialsUseListener.java
```diff
@@ -0,0 +1,77 @@
...
+    public static void fireUse(Credentials c, AbstractModelObject obj) {
+        for (CredentialsUseListener listener : all()) {
+            try {
+                listener.onUse(c, obj);
+            } catch (ThreadDeath td) {
+                throw td;
+            } catch (Throwable t) {
+                Logger.getLogger(CredentialsUseListener.class.getName()).log(Level.WARNING, null, t);
+            }
+        }
```
jglick, line 52: Try `Listeners.notify`.

src/main/java/com/cloudbees/plugins/credentials/CredentialsUseListener.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * Returns all the registered {@link CredentialsUseListener} descriptors.
+     */
+    public static ExtensionList<CredentialsUseListener> all() {
```
jglick, line 73: can be deleted

src/main/java/com/cloudbees/plugins/credentials/CredentialsUseListener.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A Listener to track {@link Credentials } usage.
+ */
+public abstract class CredentialsUseListener implements ExtensionPoint {
+
+    /**
+     * Stop listening to {@link CredentialsUseListener}.
+     */
+    public void unregister() {
```
jglick, line 20: can be deleted

src/main/java/com/cloudbees/plugins/credentials/CredentialsUseListener.java
```diff
@@ -0,0 +1,77 @@
...
+        all().remove(this);
+    }
+
+    /**
+     * Called when {@link Credentials} is used by an object, usually either {@link Run} or {@link hudson.model.Node}
+     *
+     * @param c   The used Credentials.
+     * @param obj The object using the credentials. Object is usually either {@link Run} or {@link hudson.model.Node}
+     */
+    public abstract void onUse(Credentials c, AbstractModelObject obj);
```
jglick, line 30: Why not simply `ModelObject`?

jglick:
> how to track that kind of access

Best to move `track` calls upstream into this plugin where possible. Recheck behavior after https://github.com/jenkinsci/github-branch-source-plugin/pull/527, and see my comment in #293 (this PR would be a reasonable place to fix up any missing tracking).

src/main/java/com/cloudbees/plugins/credentials/CredentialsUseListener.java
```diff
@@ -0,0 +1,29 @@
...
+
+import hudson.ExtensionPoint;
+import hudson.model.ModelObject;
+import jenkins.util.Listeners;
+
+
+/**
+ * A Listener to track {@link Credentials } usage.
+ */
+public abstract class CredentialsUseListener implements ExtensionPoint {
```
jglick, line 11: Or could be an `interface`.

src/main/java/com/cloudbees/plugins/credentials/CredentialsUseListener.java
```diff
@@ -1,77 +1,29 @@
...
-            } catch (ThreadDeath td) {
-                throw td;
-            } catch (Throwable t) {
-                Logger.getLogger(CredentialsUseListener.class.getName()).log(Level.WARNING, null, t);
-            }
-        }
+    public static void fireUse(Credentials c, ModelObject obj) {
+        Listeners.notify(CredentialsUseListener.class, true, listener -> {
+            listener.onUse(c, obj);
+        });
```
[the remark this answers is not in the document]
jglick: (in which case it would also be simple enough to just inline into call sites)

src/main/java/com/cloudbees/plugins/credentials/CredentialsUseListener.java
```diff
@@ -0,0 +1,29 @@
+package com.cloudbees.plugins.credentials;
+
+import hudson.ExtensionPoint;
+import hudson.model.ModelObject;
+import jenkins.util.Listeners;
+
+
+/**
+ * A Listener to track {@link Credentials } usage.
```
jglick, line 9:
```suggestion
 * A Listener to track {@link Credentials} usage.
```

Add `@see` links to `track` methods.

jglick (approved):
Looks fine to me. Did you successfully use this API revision in the downstream PR?

meiswjn:
Yes, the new API revision works perfectly.

jglick:
If all goes well, a new release should appear shortly.

PR #316, opened 2022-05-26
[JENKINS-68616] Enable field validation checks

MarkEWaite:
## [[JENKINS-68616](https://issues.jenkins.io/browse/JENKINS-68616)] Enable field validation checks

Regression in [1105 release](https://github.com/jenkinsci/credentials-plugin/releases/tag/1105.vb_4e24a_c78b_81).  The `div` that was added to surround the `select` seems to prevent calls to the `attrs.checkMethod`.

When using a GitHub branch source, an empty credentials field no longer displays the warning that a credential is recommended.  Other uses of credentials seem to have the same behavior.  Validation checks are no longer called.

When the `div` is included in the page, the credential validation callback is not invoked. When the `div` is removed from the page, the credential validation callback is invoked.

It is unclear to me why the addition of a `div` would change the validation behavior.

Review the differences with [white space changes](https://github.com/jenkinsci/credentials-plugin/pull/316/files?w=1) suppressed so that it is clear the difference is only in the deletion of the `div` tags that wrapped the `select`.

Not sure how to create a viable test that confirms the fix is behaving correctly.

Revert one HTML `div` from "Minor improvements"

This reverts part of commit 6dee9b8602f4ba355209f46ddd642d6daca3b448.

### DIFF
--- a/src/main/resources/lib/credentials/select.jelly
+++ b/src/main/resources/lib/credentials/select.jelly
@@ -107,8 +107,6 @@
           <input type="hidden" name="includeUser" value="${includeUser}"/>
         </j:otherwise>
       </j:choose>
-      <!-- Placing this in a div will block the field validation. The attrs.checkMethod is not called -->
-      <!-- https://issues.jenkins.io/browse/JENKINS-68616 -->
       <f:select clazz="${attrs.clazz} credentials-select" field="${attrs.field}"
                 default="${paramDefault?'':attrs.default}" value="${paramValue?(paramDefault?'':attrs.default):value}"
                 checkMethod="${attrs.checkMethod}"/>
@@ -116,7 +114,7 @@
       <j:set var="storeItems" value="${selectHelper.getStoreItems(context, includeUser)}"/>
       <j:choose>
         <j:when test="${selectHelper.hasCreatePermission(context, includeUser) and storeItems != null and !storeItems.isEmpty()}">
-          <button class="credentials-add-menu hetero-list-add jenkins-button jenkins-button--transparent" menualign="${attrs.menuAlign}"
+          <button class="credentials-add-menu hetero-list-add jenkins-button jenkins-button--transparent jenkins-!-margin-top-2" menualign="${attrs.menuAlign}"
                   suffix="${attrs.name}">
             <l:icon src="symbol-add"/>
             ${%Add}
@@ -149,7 +147,7 @@
           </div>
         </j:when>
         <j:otherwise>
-          <button class="credentials-add jenkins-button jenkins-button--transparent">
+          <button class="credentials-add jenkins-button jenkins-button--transparent jenkins-!-margin-top-2">
             <l:icon src="symbol-add"/>
             ${%Add}
           </button>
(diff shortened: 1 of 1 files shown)


MarkEWaite:
@jglick and @basil in case you are interested.

jglick:
#287 for the record

src/main/resources/lib/credentials/select.jelly
```diff
@@ -107,11 +107,11 @@
           <input type="hidden" name="includeUser" value="${includeUser}"/>
         </j:otherwise>
       </j:choose>
-      <div class="form-group jenkins-form-item jenkins-form-item--tight">
-        <f:select clazz="${attrs.clazz} credentials-select" field="${attrs.field}"
```
jglick, line 111: **Hide whitespace** to review

basil (approved):

src/main/resources/lib/credentials/select.jelly
MarkEWaite: > **Hide whitespace** to review

Agreed.  Much clearer with [white space diffs suppressed](https://github.com/jenkinsci/credentials-plugin/pull/316/files?w=1).

jglick:
(I have merge/release permissions but would prefer to get an opinion from @timja as author of #287, since I know little about this subject.)

MarkEWaite:
> (I have merge/release permissions but would prefer to get an opinion from @timja as author of #287, since I know little about this subject.)

I agree that we should let this wait until @timja has had time to reply.  The missing validation is unfriendly for the user but does not break functionality as far as I know.

timja:
Looking

src/main/resources/lib/credentials/select.jelly
```diff
@@ -107,11 +107,11 @@
...
-                  default="${paramDefault?'':attrs.default}" value="${paramValue?(paramDefault?'':attrs.default):value}"
-                  checkMethod="${attrs.checkMethod}"/>
-      </div>
+      <!-- Placing this in a div will block the field validation. The attrs.checkMethod is not called -->
```
timja, line 110: this is not correct, it's one of the classes, anyway I'll push something to fix it up

jglick:
All Greek to me. @MarkEWaite if you can confirm the last edit preserves the bug fix and general appearance, I can get a release cut.

MarkEWaite:
> All Greek to me. @MarkEWaite if you can confirm the last edit preserves the bug fix and general appearance, I can get a release cut.

Confirmed that the fix the most recent change preserves the bug fix and the general appearance.

basil (approved):
I support the merge and release of this fix, but I would prefer it if the Jira ticket be updated with an evaluation of the cause of the problem. This project does not have much of a culture of writing root cause analysis in bugs, so this is an optional comment, but I believe that preserving a clear and written analysis of the cause of the problem, including a description of how the new code addresses the problem (beyond just "fix spacing") would benefit everyone involved. This is optional, just for your consideration.

timja:
@MarkEWaite fix re-introduced the UI bug of elements touching each other.
Fix spacing, fixed the spacing so they weren’t touching.

I didn’t attempt to track down a root cause, likely due to some tree building in hudson-behaviour.js which one of the classes was affecting.

timja:
I’m unsure @janfaracik may know

basil:
There is nothing wrong with being unsure when starting out, but the goal should be to become sure through analytical reasoning, regardless of whether or not one was the original author of the code.

PR #327, opened 2022-06-21
Add snapshot taker for special `StandardUsernamePasswordCredentials` impls

jamesrobson-secondmind:
This adds a snapshot taker for StandardUsernamePasswordCredentials.  This is needed to fix an issue in hashicorp-vault plugin which currently prevents ssh keys being sent to agents which is being developed in https://github.com/jenkinsci/hashicorp-vault-plugin/pull/218

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsSnapshotTaker.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsSnapshotTaker.java
@@ -25,6 +25,8 @@ public class UsernamePasswordCredentialsSnapshotTaker extends CredentialsSnapsho
         if (credentials instanceof UsernamePasswordCredentialsImpl) {
             return credentials;
         }
-        return new UsernamePasswordCredentialsImpl(credentials.getScope(), credentials.getId(), credentials.getDescription(), credentials.getUsername(), Secret.toString(credentials.getPassword()));
+        UsernamePasswordCredentialsImpl snapshot = new UsernamePasswordCredentialsImpl(credentials.getScope(), credentials.getId(), credentials.getDescription(), credentials.getUsername(), Secret.toString(credentials.getPassword()));
+        snapshot.setUsernameSecret(credentials.isUsernameSecret());
+        return snapshot;
     }
 }
--- a/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsSnapshotTaker.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsSnapshotTaker.java
@@ -1,0 +1,27 @@
+package com.cloudbees.plugins.credentials.impl;
+
+import com.cloudbees.plugins.credentials.CredentialsSnapshotTaker;
+import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
+
+import hudson.Extension;
+import hudson.util.Secret;
+
+@Extension
+public class UsernamePasswordCredentialsSnapshotTaker extends CredentialsSnapshotTaker<StandardUsernamePasswordCredentials> {
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Class<StandardUsernamePasswordCredentials> type() {
+        return StandardUsernamePasswordCredentials.class;
+    }
+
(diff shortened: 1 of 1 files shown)


jetersen (approved):

jetersen:
cc @jglick

jglick (approved):
Looks right.

jglick (changes requested):
Ah no, it fails to copy the `usernameSecret` field.

I would suggest just returning the argument unchanged if it happens to be a `UsernamePasswordCredentialsImpl`.

src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsSnapshotTaker.java
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public StandardUsernamePasswordCredentials snapshot(StandardUsernamePasswordCredentials credentials) {
+        if (credentials instanceof UsernamePasswordCredentialsImpl) {
+            return credentials;
+        }
+        return new UsernamePasswordCredentialsImpl(credentials.getScope(), credentials.getId(), credentials.getDescription(), credentials.getUsername(), Secret.toString(credentials.getPassword()));
```
jglick, line 28:
```suggestion
        UsernamePasswordCredentialsImpl snapshot = new UsernamePasswordCredentialsImpl(credentials.getScope(), credentials.getId(), credentials.getDescription(), credentials.getUsername(), Secret.toString(credentials.getPassword()));
        snapshot.setUsernameSecret(credentials.isUsernameSecret());
        return snapshot;
```
just in case other impls now implement this parameter as well.

jtnord (approved):

jamesrobson-secondmind:
We have a confirmation of this change working with vault (https://github.com/jenkinsci/hashicorp-vault-plugin/issues/201#issuecomment-1164366081), from commit `e331be4`. I've tested the following changes my self so we should now be good to go.

jetersen (approved):

jglick (approved):

jtnord:
noting this seemingly breaks `GitHubAppCredentials` as it claims support for them and snapshots them 😱 

```
org.jenkinsci.plugins.github_branch_source.GitHubAppCredentials.generateAppInstallationToken(GitHubAppCredentials.java:216)
  at org.jenkinsci.plugins.github_branch_source.GitHubAppCredentials.getToken(GitHubAppCredentials.java:298)
  at org.jenkinsci.plugins.github_branch_source.GitHubAppCredentials.getPassword(GitHubAppCredentials.java:327)
  at com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsSnapshotTaker.snapshot(UsernamePasswordCredentialsSnapshotTaker.java:28)
  at com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsSnapshotTaker.snapshot(UsernamePasswordCredentialsSnapshotTaker.java:9)
  at com.cloudbees.plugins.credentials.CredentialsProvider.snapshot(CredentialsProvider.java:821)
```

jglick:
@jtnord by “breaks” do you mean specifically the token refresh https://github.com/jenkinsci/github-branch-source-plugin/blob/7618247e672d2008fa8dc006ee8e87f4e64ab2c4/src/main/java/org/jenkinsci/plugins/github_branch_source/GitHubAppCredentials.java#L497?

jtnord:
> @jtnord by “breaks” do you mean specifically the token refresh https://github.com/jenkinsci/github-branch-source-plugin/blob/7618247e672d2008fa8dc006ee8e87f4e64ab2c4/src/main/java/org/jenkinsci/plugins/github_branch_source/GitHubAppCredentials.java#L497?
> 
> If so, I guess the correct fix is to replace `writeReplace` with a new `CredentialsSnapshotTaker` taking precedence over `UsernamePasswordCredentialsSnapshotTaker`.
> 
> https://github.com/jenkinsci/credentials-plugin/blob/be8bbceee34777d4dcae5ae1597aa32d775b0a5f/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java#L808-L814
> 
> should make this straightforward: no need for https://javadoc.jenkins.io/hudson/Extension.html#ordinal()

I'm working on it.  there be some dragons I am trying to understand first

PR #364, opened 2022-09-04
Modernise

timja:
- Use context specific sidepanel
- Remove sidepanel on main credentials page
- Display store specific names rather than `Jenkins` for all of them on view credentials page
- Remove `norefresh` from views
- Maintain `manage` in url bar so that breadcrumbs work correctly

_Note: screenshots taken on 2.366, on 2.361 the `one-column` layouts will be a bit wider and not look quite as good_

<table>
    <thead>
    <tr>
        <th>
        <th>Before
        <th>After
    </tr>
    <thead>
    <tr>
        <td>View Credentials
        <td>image
        <td>image
    <tr>
        <td>View domain (one-column)
        <td>image
        <td>image
    </tr>
    <tr>
        <td>View domain (domain modifiable)
        <td>image
        <td>image
   </tr>
    <tr>
        <td>View store (domain not modifiable)
        <td>image
        <td>image
    </tr>
    <tr>
        <td>View store (domain modifiable)
        <td>image
        <td>image
    </tr>
    <tr>
    <td>Add domain
     <td>image
      <td>image
     </tr>
</table>

TODO:
- [x] Remove some duplication between `ViewCredentialsAction` and `CredentialsStoreAction`~
- ~Look at fixing [JENKINS-68791](https://issues.jenkins.io/browse/JENKINS-68791) probably with https://github.com/jenkinsci/jenkins/pull/6659~ - will do separately
- [x] Review folder plugin with its credentials provider
- [x] Before and after screenshots


src/main/resources/com/cloudbees/plugins/credentials/ViewCredentialsAction/index.properties
```diff
@@ -1,2 +1,2 @@
 global=(global)
-storesScopedTo=Stores scoped to <a href=".." class='model-link inside'>{0}</a>
+storesScopedTo=Stores scoped to {0}
```
daniel-beck, line 2: Why?

src/main/java/com/cloudbees/plugins/credentials/ViewCredentialsAction.java
```diff
@@ -58,6 +58,7 @@
 import jenkins.model.ModelObjectWithContextMenu;
 import jenkins.model.TransientActionFactory;
 import org.acegisecurity.Authentication;
+import org.apache.commons.lang.StringUtils;
```
daniel-beck, line 61: Unused

src/main/resources/com/cloudbees/plugins/credentials/ViewCredentialsAction/index.properties
timja: It just didn't seem useful, why would you want that there?

I looked at renaming this to be clearer what it was but I didn't come up with anything

daniel-beck: Good point, I missed that this only ever links to the current object, not to its ancestors (which might have been useful).

daniel-beck:
Looks reasonable overall. Especially thanks for taking care of the confusing sidepanels :)

pom.xml
```diff
@@ -67,7 +67,7 @@
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
-    <jenkins.version>2.340</jenkins.version>
+    <jenkins.version>2.361</jenkins.version>
```
timja, line 70: required for the "Manage Jenkins" breadcrumb to be added.

src/main/java/com/cloudbees/plugins/credentials/CredentialsStore.java
```diff
@@ -537,6 +538,22 @@ public final String getContextDisplayName() {
         }
     }
 
+    /**
+     * Resolves a display name from the Store
+     * @return the display name
+     */
+    public String getDisplayName() {
```
timja, line 545: Moved code from `CredentialsStoreAction`

janfaracik:
Looks good to me!

---



For this page, how would you feel about moving the 'Add domain' into the app bar and making it a primary button? (like the System Log page)

Screenshot 2022-09-06 at 20 07 02

timja:
Yep will try it, it was looking quite weird with a single item there

timja:
mind taking another look? I'll update screenshots later on

pom.xml
timja: yes tomorrow I will update

NotMyFault (approved):

janfaracik (approved):

pom.xml
timja:
```suggestion
    <jenkins.version>2.361.1</jenkins.version>
```

timja:
@jenkinsci/credentials-plugin-developers possible for a review please?

jtnord:
when I go to add a credential I get a page (2 column) that seems out of place with the new design

Given I am in the "add credential" of a current domain" should the "confgure domain" delete/domain and "add credential" show up here?
image

jtnord (changes requested):
something is definatly wrong with the delete button, but see more feedback above

(is red with 2.688 but not with the LTS).

timja:
all the feedback should be addressed 

the class was renamed in 2.364 I could add both, but some of the pages look a bit funny on a one-column layout without other changes in 2.364 so I think better to just raise the baseline to that.

> When viewing Folder credentials there is no obvious way to guide you where to go. (yes the existing design is shitty) but there was a subtask under the credentials task which highlighted the "folder"

I have potential future plans to remove the stores from that page, add a search bar and create a new create credential experience where you can choose the store / type / context in a nicer page.

but I think out of scope of this PR and this should be good enough for now?

jtnord:
> the class was renamed in 2.364 I could add both, but some of the pages look a bit funny on a one-column layout without other changes in 2.364 so I think better to just raise the baseline to that.

I'm a bit hesitant on bumping the baseline this early in the LTS cycle. (There's some fixes still needed to be written for the icon change) and I would like to avoid branching if possible.

Additionally I thought the pages whilst not perfect in the LTS where still an improvement over the previous code.

Would you be against dropping the version requirement to the LTS and adding the old class name in addition?

timja:
Done

jtnord (approved):

jtnord:
Thanks @timja

ite-klass:
A more descriptive PR title would have been helpful for the release changelog

ite-klass:
*Modernize UI* would clarify it is about user interface / presentation changes

PR #365, opened 2022-09-08
Move app-bar to main panel

timja:
Requires https://github.com/jenkinsci/jenkins/pull/7051

App-bar is getting a new feature to allow it to be in the sidepanel, it's now required to place it in the section you want it displayed.

_Note: no compatible way of doing this currently, it needs to be released when core is released, given there's only 3 plugins using it this should be fine_


jtnord:
obsolete by #366

PR #366, opened 2022-09-08
Move app-bar to main panel

timja:
downstream of https://github.com/jenkinsci/jenkins/pull/7051 (released in 2.374)

App-bar has received a new feature to allow it to be in the sidepanel, it's now required to place it in the section you want it displayed.

Currently there's a minor UI issue with credentials plugin on 2.374 without this as the headings will be in the wrong place right now


jimklimov:
hindsight-nit: isn't "update to LTS branch" a misnomer when LTS is 2.361.4 and update is to 2.375?
As a result, rebuilds can't be installed to the currently GA LTS version... (locally, reverting the numbers seems to have worked)

timja:
You don’t need this on old versions. The LTS version reference is for the next line being released on the 30th

jimklimov:
i needed current plugin on older version (current LTS)
well, sort of needed - to deploy my PR based against master

timja:
> i needed current plugin on older version (current LTS)
> well, sort of needed - to deploy my PR based against master

Right best to either adjust locally and build or create a separate PR that’s not for merge that has the changes necessary to run on that version

jimklimov:
(Ab)using PRs to build the incremental version? :)

Or just to rebase PRed changes over older baseline? (the latter feels clumsy, in case the PR needs to evolve so lots of dual-cherrypicking involved).

Anyhow, originally it was just a nit about "update to LTS branch" not being really that at that moment, making the result not usable for currently published LTS.

PR #373, opened 2022-09-16
Remove useless broken "Add" button on credentials-select-control.

jamie-pate:
I don't know if there is a way to get a 'data-url' into the template to actually fix this button.

See this change that broke it: https://github.com/jenkinsci/credentials-plugin/commit/9c5bb72106824ea826ba0d5e4d63fe5d2a11aa91#diff-45693b8474f3341b5d603b463cd4c10f4ec301a1c8c1c89b67034ba81056dfc9R44

- [x ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x ] Ensure that the pull request title represents the desired changelog entry
- [x ] Please describe what you did
- [? ] Link to relevant issues in GitHub or Jira (there is no issues tab on github, and where is this Jira?)
- [x ] Link to relevant pull requests, esp. upstream and downstream changes
- [? ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/resources/lib/credentials/select/select.js
+++ b/src/main/resources/lib/credentials/select/select.js
@@ -214,10 +214,7 @@ Behaviour.specify("BUTTON.credentials-add-menu", 'credentials-select', -99, func
     }
     e=null;
 });
-Behaviour.specify("BUTTON.credentials-add", 'credentials-select', 0, function (e) {
-    makeButton(e, e.disabled ? null : window.credentials.add);
-    e = null; // avoid memory leak
-});
+
 Behaviour.specify("DIV.credentials-select-control", 'credentials-select', 100, function (d) {
     d = $(d);
     var buttons = d.getElementsBySelector("INPUT.credentials-select-radio-control");


jtnord:
can you describe in detail what doesn;t work or why it is broken?  Perhaps a Jira issue?

jamie-pate:
It looks like it was just left out in a refactor but never removed... It just does nothing

jtnord:
https://github.com/jenkinsci/credentials-plugin/blob/3f165e9abbdf4c608f86eea2ceb9e0d53c068ef0/src/main/resources/lib/credentials/select.jelly#L151-L154  ?

jglick:
I think you are right that it is broken, but also that your real problem may be https://issues.jenkins.io/browse/JENKINS-19413?focusedId=456506&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-456506

timja:
Going to close this, the button is definitely working

PR #377, opened 2022-10-11
[JENKINS-68791] Missing icon for global credential store

ridemountainpig:
See [JENKINS-68791](https://issues.jenkins.io/browse/JENKINS-68791).
### Before
image

### After
image

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/SystemCredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/SystemCredentialsProvider.java
@@ -629,7 +629,7 @@ public class SystemCredentialsProvider extends AbstractDescribableImpl<SystemCre
         @Override
         public String getIconFileName() {
             return isVisible()
-                    ? "/jenkins/plugin/credentials/images/system-store.svg"
+                    ? "/plugin/credentials/images/system-store.svg"
                     : null;
         }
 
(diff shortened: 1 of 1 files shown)


ridemountainpig:
@daniel-beck can you help me review this PR ? Thanks ~

src/main/java/com/cloudbees/plugins/credentials/SystemCredentialsProvider.java
```diff
@@ -629,7 +629,7 @@ public CredentialsStore getStore() {
         @Override
         public String getIconFileName() {
             return isVisible()
-                    ? "symbol-jenkins"
+                    ? "/jenkins/plugin/credentials/images/system-store.svg"
```
jtnord, line 632: I think this will only work if you have Jenkins deployed in the `jenkins` context.

also this should be using the [icon framework](https://github.com/jenkinsci/credentials-plugin/blob/de334b12145ca0c1eeb19689246fdc58fa3c0ca7/src/main/java/com/cloudbees/plugins/credentials/ContextMenuIconUtils.java#L180-L203) to lookup the icon.

so I think you just need

```suggestion
                    ? "system-store"
```

jtnord: icon class is speficied in getIconClass not Filename.  this method appears to be totally unused if the class is present and represents an icon.

jtnord (approved):

PR #385, opened 2022-11-02
Adapt to https://github.com/jenkinsci/jenkins/pull/7293

basil:
Now that core includes `antlr4-runtime` as of https://github.com/jenkinsci/jenkins/pull/7293, there is no need to depend on this from `credentials-plugin`. Unfortunately pending [MNG-5588](https://issues.apache.org/jira/browse/MNG-5588) we can't get the version of the `antlr4-maven-plugin` from the core BOM so this has to be kept in sync manually.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,8 +67,7 @@
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
-    <jenkins.version>2.361.1</jenkins.version>
-    <antlr4.version>4.11.1</antlr4.version>
+    <jenkins.version>2.376</jenkins.version>
   </properties>
 
   <repositories>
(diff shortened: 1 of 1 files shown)


jtnord (commented):
Thanks @basil .

As this requires a new core version that will not be an LTS for ~ 3 months (2.375 was chosen for the next LTS) do you have any objection if we just leave this here for a while.

I don't really want to branch to have any bug fixes/improvements land for that amount of time and IIUC the plugin will still work in newer Jenkins cores, it is just that this is a cleanup.

basil:
> do you have any objection if we just leave this here for a while.

Yes this is just a cleanup, and I have no objection if we leave this here for a maximum of six (6) months.

basil:
> I have no objection if we leave this here for a maximum of six (6) months.

Six (6) months will have elapsed in less than a week.

PR #411, opened 2023-01-24
feat: add merge support for casc defined system credentials

cronik:
Enables support for merging casc defined credentials with existing credentials (i.e. manually created).

In some environments it is desirable to define some credentials declaratively while also be able to define credentials through the UI (which out otherwise be defined in plain text or in encoded format which could be easily transferable).

One such example is mixing vault credentials defined in casc config with those defined directly through Jenkins. In the below casc example it is assumed the the `vault-approle` credential was created through the Jenkins UI. Currently this is not possible since the casc system credentials will remove all non-casc defined credentials on restart.

```yaml
unclassified:
  hashicorpVault:
    configuration:
      vaultCredentialId: "vault-approle"
      vaultUrl: "https://myvault.mycorp.net"
credentials:
  system:
    domainCredentials:
      - credentials:
          - vaultUsernamePasswordCredentialImpl:
              description: "vault managed credential"
              id: "my-vault-cred"
              path: "kv/jenkins"
              scope: GLOBAL
```

Hoping to get some feedback on best approach to implement this feature before finalizing the PR. I have 2 approaches that could work:
1. Use a env var and/or system property to enabled merge versus override
2. define an explicit casc attribute for merging credentials (i.e. `mergeDomainCredentials`)

fixes JENKINS-64079


src/main/java/com/cloudbees/plugins/credentials/casc/SystemCredentialsProviderConfigurator.java
```diff
@@ -64,17 +75,29 @@ protected SystemCredentialsProvider instance(Mapping mapping, ConfigurationConte
     public Set<Attribute<SystemCredentialsProvider, ?>> describe() {
         return Collections.singleton(
             new MultivaluedAttribute<SystemCredentialsProvider, DomainCredentials>("domainCredentials", DomainCredentials.class)
-                .setter( (target, value) -> target.setDomainCredentialsMap(DomainCredentials.asMap(value)))
+                .setter((target, value) -> {
+                    String strategy = Util.fixEmptyAndTrim(
+                            System.getProperty("casc.credentials.merge.strategy",
+                            System.getenv("CASC_CREDENTIALS_MERGE_STRATEGY")
+                    ));
```
jtnord, line 82: rather than put the strategy in an environment variable - is it not possible to manage it purely as code in the yaml file?

mikecirioli: I'd go even further and define the default strategy value as `replace` to make it explicit.  Example yaml might look like:

```
- MergeStrategies
    - CascCredentials: merge
````

**Comment by cronik** [comment]
> If I understand this correctly it means you can only ever add or update credentials via CasC. If you added one via CasC it can only be removed via the UI (as once persisted there is no way to tell the difference)?
> 
> > Currently this is not possible since the casc system credentials will remove all non-casc defined credentials on restart.
> 
> That is kind of the point of CasC though, you are managing your instance via code - not via the UI?

I agree the behavior introduced is not in the true spirit of CasC being the source of config. If the merge strategy is configured then it would require manual removal of credentials that were added via CasC. 

In our Jenkins deployments we are trying to get teams to transition to vault credentials configured via CasC. Some credentials would be manually created while others are defined in CasC during this transition. We would accept the extra work of having to manually remove items no longer defined in CasC.

Would something like a CasC specific CredentialStore implementation be a better solution? I'm thinking along the lines of something similar to [KubernetesCredentialStore](src/main/java/com/cloudbees/jenkins/plugins/kubernetes_credentials_provider/KubernetesCredentialsStore.java). This would mean:

- `cascCredentials` could not be added/removed/updated via the UI
- Removed items in `cascCredentials` would be reflected without manual intervention
- Credentials added via the UI would not be impacted since they would be managed under `domainCredentials`

```yaml
credentials:
  system:
    domainCredentials:
      - credentials:
          - usernamePasswordCredential: {id: foo}
    cascCredentials: 
      - credentials:
          - usernamePasswordCredential: {id: bar}
```

lasfarnabil:
Hello, We want to use credentials-plugins with casc declaration in order to specify default credentials of jenkins instance. And we need to keep to users the ability to add credentials also in global system creds. For this reason, we need this feature to be added to next release if possible ? 
we need to have this option to merge credentials defined by casc and credentials defined by UI.

jtnord:
> Hello, We want to use credentials-plugins with casc declaration in order to specify default credentials of jenkins instance. And we need to keep to users the ability to add credentials also in global system creds. For this reason, we need this feature to be added to next release if possible ? we need to have this option to merge credentials defined by casc and credentials defined by UI.

This PR will not  be merged as is given the outstanding comments (use of a property or env var).

If this is blocking you then feel free to create an updated PR based on this.

PR #481, opened 2023-09-13
Do not use `FormApply#applyResponse` to execute arbitrary javascript

yaroslavafenkin:
Motivated by https://github.com/jenkinsci/jenkins/pull/8351.
That PR changes `FormApply#applyResponse` method to make it CSP compatible in a way that does not allow to execute arbitrary javascript as a callback.

Looking at the usages of `FormApply#applyResponse` in the ecosystem its most common (and perhaps intended) use is to show a notification after "Apply" button (or alike) was clicked. In credentials-plugin it's used in a "add credentials in-place" form, and the calls are rather uncommon.
Two calls are `FormApply.applyResponse("window.alert(...)")`, which could easily be replaced by the new CSP compatible version `FormApply.showNotification` from the core PR linked above.
Third call is worse - `FormApply.applyResponse("window.credentials.refreshAll();")`.
My initial attempt was to keep using `FormApply.applyResponse` mechanism, but moving `window.credentials.refreshAll()` to .js and making that execute at the right time turned out to be far from trivial.

After analysing the code for a bit more there seems to be no reason to keep using `FormApply.applyResponse`, hence this PR gets rid of it. It changes the way that form is submitted, and gets rid of javascript code needed for the `FormApply.applyResponse` mechanism to work. Instead server now responds with data for the notification, and javascript code uses that data to call `notificationBar.show(...)`. `window.credentials.refreshAll()` was also moved to .js file, and is executed after credentials were successfully added and response was received.

### Testing done
Tested interactively by adding credentials via "add credentials in-place" form. Made sure credentials were added successfully and immediately available in the dropdown after the form popup was closed.
Tested error paths by attaching the debugger and changing values through that, since I found no way to make credentials domain non-modifiable via UI.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
@@ -600,8 +600,8 @@ public class CredentialsSelectHelper extends Descriptor<CredentialsSelectHelper>
             if (!store.isDomainsModifiable()) {
                 hudson.util.HttpResponses.status(400).generateResponse(req, rsp, null);
                 return new JSONObject()
-                        .accumulate("message", "Domain is read-only")
-                        .accumulate("notificationType", "ERROR");
+                        .element("message", "Domain is read-only")
+                        .element("notificationType", "ERROR");
             }
             store.checkPermission(CredentialsStoreAction.CREATE);
             JSONObject data = req.getSubmittedForm();
@@ -610,15 +610,22 @@ public class CredentialsSelectHelper extends Descriptor<CredentialsSelectHelper>
             if (!store.getDomains().contains(wrapper.getDomain())) {
                 hudson.util.HttpResponses.status(400).generateResponse(req, rsp, null);
                 return new JSONObject()
-                        .accumulate("message", "Store does not have selected domain")
-                        .accumulate("notificationType", "ERROR");
+                        .element("message", "Store does not have selected domain")
+                        .element("notificationType", "ERROR");
             }
             store.checkPermission(CredentialsStoreAction.CREATE);
             Credentials credentials = req.bindJSON(Credentials.class, data.getJSONObject("credentials"));
-            store.addCredentials(wrapper.getDomain(), credentials);
-            return new JSONObject()
-                    .accumulate("message", "Credentials created")
-                    .accumulate("notificationType", "SUCCESS");
+            boolean credentialsWereAdded = store.addCredentials(wrapper.getDomain(), credentials);
+            if (credentialsWereAdded) {
+                return new JSONObject()
+                        .element("message", "Credentials created")
+                        .element("notificationType", "SUCCESS");
+            } else {
+                return new JSONObject()
+                        .element("message", "Credentials with specified ID already exist in " + domainName + " domain")
+                        // TODO: or domain does not exist at all?
+                        .element("notificationType", "ERROR");
+            }
         }
 
         /**
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
@@ -38,7 +38,6 @@ import hudson.model.ModelObject;
 import hudson.model.User;
 import hudson.security.AccessControlled;
 import hudson.security.Permission;
-import hudson.util.FormApply;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
```diff
@@ -596,24 +595,30 @@ public WrappedCredentialsStore(@NonNull ContextResolver resolver, @NonNull Crede
...
+                return new JSONObject()
+                        .accumulate("message", "Store does not have selected domain")
+                        .accumulate("notificationType", "ERROR");
             }
             store.checkPermission(CredentialsStoreAction.CREATE);
             Credentials credentials = req.bindJSON(Credentials.class, data.getJSONObject("credentials"));
             store.addCredentials(wrapper.getDomain(), credentials);
-            FormApply.applyResponse("window.credentials.refreshAll();").generateResponse(req, rsp, null);
+            return new JSONObject()
+                    .accumulate("message", "Credentials created")
```
Kevin-CB, line 620: Well, that doesn't always result in the creation of credentials, e.g when ID already exists, it won't create anything.
You could use the boolean result of `addCredentials(...)` to adapt your message.

Kevin-CB (approved):
Except for the small comment I made, I have nothing to add.
I tested your PR locally and it works as expected.

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
yaroslavafenkin: > You could use the boolean result of addCredentials(...) to adapt your message.

Oh, that sounds great to me, I've missed that there's return value from `addCredentials`.

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
```diff
@@ -596,24 +595,30 @@ public WrappedCredentialsStore(@NonNull ContextResolver resolver, @NonNull Crede
...
         @RequirePOST
-        public void doAddCredentials(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException {
+        @Restricted(NoExternalUse.class)
+        public JSONObject doAddCredentials(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException {
             if (!store.isDomainsModifiable()) {
                 hudson.util.HttpResponses.status(400).generateResponse(req, rsp, null);
-                FormApply.applyResponse("window.alert('Domain is read-only')").generateResponse(req, rsp, null);
+                return new JSONObject()
+                        .accumulate("message", "Domain is read-only")
+                        .accumulate("notificationType", "ERROR");
```
daniel-beck, line 604: Any reason to use this over `#element`? We don't want the accumulate special behavior after all.

daniel-beck (approved):

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
yaroslavafenkin: > Any reason to use this over #element?

No particular reason, I wasn't sure which one was better to use.

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
yaroslavafenkin: I think it would be weird to show a notification saying "Credentials not added" and not specifying reason why. So I'm unsure at the moment how to act on this, will give it another thought.

yaroslavafenkin: And also if domain doesn't exist it seems. See https://github.com/jenkinsci/credentials-plugin/blob/860724bea804fb39b0967f4d3a528600e399147c/src/main/java/com/cloudbees/plugins/credentials/SystemCredentialsProvider.java#L283 Unsure under what circumstances that path would be reachable.

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
yaroslavafenkin: https://github.com/jenkinsci/credentials-plugin/pull/481/commits/d71602e7cc03a96bf876171da18a19d8dcb7efb9

daniel-beck (approved):

car-roll (approved):
LGTM but to be fair, I focused more on the java part

rsandell (approved):

PR #502, opened 2024-01-25
[JENKINS-72611] Forbid editing credentials ID

yaroslavafenkin:
https://issues.jenkins.io/browse/JENKINS-72611

### Testing done
Tested interactively, made sure that two credentials cannot have the same ID.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/SystemCredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/SystemCredentialsProvider.java
@@ -323,6 +323,14 @@ public class SystemCredentialsProvider extends AbstractDescribableImpl<SystemCre
     private synchronized boolean updateCredentials(@NonNull Domain domain, @NonNull Credentials current,
                                                    @NonNull Credentials replacement) throws IOException {
         checkPermission(CredentialsProvider.UPDATE);
+
+        // See IdCredentials.Helper#equals.
+        // Note: We probably should not assume that all credentials are IdCredentials. We also do this a few lines below in list.indexOf(current) call.
+        // If this one breaks, that one breaks as well.
+        if (!current.equals(replacement)) {
+            throw new IllegalArgumentException("Credentials' IDs do not match, will not update.");
+        }
+
         Map<Domain, List<Credentials>> domainCredentialsMap = getDomainCredentialsMap();
         if (domainCredentialsMap.containsKey(domain)) {
             List<Credentials> list = domainCredentialsMap.get(domain);
--- a/src/main/java/com/cloudbees/plugins/credentials/UserCredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/UserCredentialsProvider.java
@@ -392,6 +392,14 @@ public class UserCredentialsProvider extends CredentialsProvider {
         private synchronized boolean updateCredentials(@NonNull Domain domain, @NonNull Credentials current,
                                                        @NonNull Credentials replacement) throws IOException {
             checkPermission(CredentialsProvider.UPDATE);
+
+            // See IdCredentials.Helper#equals.
+            // Note: We probably should not assume that all credentials are IdCredentials. We also do this a few lines below in list.indexOf(current) call.
+            // If this one breaks, that one breaks as well.
+            if (!current.equals(replacement)) {
+                throw new IllegalArgumentException("Credentials' IDs do not match, will not update.");
+            }
+
             Map<Domain, List<Credentials>> domainCredentialsMap = getDomainCredentialsMap();
             if (domainCredentialsMap.containsKey(domain)) {
                 List<Credentials> list = domainCredentialsMap.get(domain);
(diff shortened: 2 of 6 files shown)


src/main/java/com/cloudbees/plugins/credentials/SystemCredentialsProvider.java
```diff
@@ -323,6 +323,14 @@ private synchronized boolean removeCredentials(@NonNull Domain domain, @NonNull
     private synchronized boolean updateCredentials(@NonNull Domain domain, @NonNull Credentials current,
                                                    @NonNull Credentials replacement) throws IOException {
         checkPermission(CredentialsProvider.UPDATE);
+
+        // See IdCredentials.Helper#equals.
```
jglick, line 327: FTR https://github.com/jenkinsci/credentials-plugin/blob/cf0a900b37c29359c89e0d85fdc48361ef4c0eb1/src/main/java/com/cloudbees/plugins/credentials/common/IdCredentials.java#L50-L85 https://github.com/jenkinsci/credentials-plugin/blob/cf0a900b37c29359c89e0d85fdc48361ef4c0eb1/src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java#L116-L130

jglick (approved):

alecharp:
Hello, this created a compatibility issue with `plain-credentials` plugin.
Since this was added, the `BaseTest` of `plain-credentials` are failing.

src/main/java/com/cloudbees/plugins/credentials/SystemCredentialsProvider.java
```diff
@@ -323,6 +323,14 @@ private synchronized boolean removeCredentials(@NonNull Domain domain, @NonNull
...
+        // Note: We probably should not assume that all credentials are IdCredentials. We also do this a few lines below in list.indexOf(current) call.
+        // If this one breaks, that one breaks as well.
```
jtnord, line 329: how so - the list is of type `Credentials` not `IdCredentials` so that would not break with a big bad angry jenkins.
At worst it would not find `current`, as the list would use `equals` and it is up to current credential may have been correctly obtained by it's name or otherwise in the code path.

jtnord:
indeed this causes regressions

-> 
image


failed to update a description of a credential.  LoadFocus.com API key

jtnord:
> @jtnord What concrete `Credentials` implementations are not `instanceof IdCredentials`? 

for example -> https://github.com/jenkinsci/loadfocus-jmeter-load-testing-integration-plugin/blob/d2d2af5b0af276932dfb6c8191d3bb13791d6bc7/src/main/java/com/loadfocus/jenkins/impl/LoadCredentialImpl.java  there may be companies with other implementations.

> I was under the impression that was essentially impossible 

Nope this works :) 

>and it was only a technicality that `getId` had not been pushed up into `Credentials` itself. (How would you even select credentials without an `id`?)

https://github.com/jenkinsci/credentials-plugin/blob/7eb51b3ac97b88bc3cdb2c62b91cf7811255bdbc/src/main/java/com/cloudbees/plugins/credentials/CredentialsStoreAction.java#L744-L748

you get a dummy ID based on the index in the list.  For add and delete.   
For usage you may be thinking too much about credentials-binding.

A plugin can perfectly ask for any credential of Type X for domain F and get the list `CredentialsProvider.lookupCredentials` and then do whatever it wants.  (expect there is only one and use the first, or use a custom different method of identifying the suitable one)

IDs are the future (or the now) but they are not /currently/ mandatory. 

> > the `BaseTest` of `plain-credentials` are failing
> 
> https://github.com/jenkinsci/plain-credentials-plugin/blob/ce8710821f17a043d91b448cf19b8568935d0167/src/test/java/org/jenkinsci/plugins/plaincredentials/BaseTest.java#L107? That just sounds like a meaningless test.

jglick:
Hmm https://github.com/jenkinsci/loadfocus-jmeter-load-testing-integration-plugin/blob/d2d2af5b0af276932dfb6c8191d3bb13791d6bc7/src/main/java/com/loadfocus/jenkins/LoadCredential.java#L13 looks like it was meant to extend `IdCredentials`. Is that the source of the screenshot above?

> expect there is only one and use the first

Are there some examples of this? I can only think of cases where config includes a `credentialsId`.

:shrug: anyway I guess it should be simple enough to fix things up to just ignore non-`IdCredentials`. Somewhere we should put a compile-time and/or runtime warning about these, because the intent was certainly for all implementations to be `IdCredentials`.

jtnord:
> I guess it should be simple enough to fix things up to just ignore non-IdCredentials

yes - we only need to do the comparison if one or other of the credentials is an `IDCredential`.

> Somewhere we should put a compile-time and/or runtime warning about these, because the intent was certainly for all implementations to be IdCredentials.

if we want to kill non ID based credentials and cleanup the code then I am ok with doing that in a PR that is clearly saying "remove support for obsolete implementations of Credentials" so long as it is just doing that, and we also fixup the knwon fallout (which is not huge but there is some).
I'm not ok with killing it by accident and without release notes or an upgrade path.

PR #540, opened 2024-06-21
Remove FileOnMasterKeyStroreSource and SECURITY-1322 migration

jtnord:
It has been 5 years since the security fix was introduced and the migration code occurred.

Whilst an admin could have still (ab)used the CLI/REST to set a FileOnMaster they could not do this from the UI creating a disparity.

This change removes the migration and the ability to use this via CLI/REST

As noted originally and as would be present in the Jenkins logs at `INFO` level: 
 `SECURITY-1322: Migrating FileOnMasterKeyStoreSource to UploadedKeyStoreSource. The containing item may need to be saved to complete the migration.` 
Users would be required to check that they have saved any jobs where migration has occurred before updating to ensure that the CertificateCredential is not lost.

### Testing done

`mvn verify`

### Submitter checklist


pom.xml
```diff
@@ -169,6 +169,9 @@
             </goals>
           </execution>
         </executions>
+        <configuration>
+          <compatibleSinceVersion>1344</compatibleSinceVersion>
```
jtnord, line 173: will likely need to be updated before the release.

src/test/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImplTest.java
```diff
@@ -113,101 +113,6 @@ public void displayName() throws IOException {
         assertEquals(EXPECTED_DISPLAY_NAME, CredentialsNameProvider.name(new CertificateCredentialsImpl(null, "abc123", null, "password", storeSource)));
     }
 
-    @Test
```
jtnord, line 116: the code these tests where testing has been deleted.
Whilst we could leave a test that shows the behaviour doesn't work, I don't see a point in that)

dwnusbaum:
@jtnord Maybe you already know, but just in case, a few plugins use `FileOnMasterKeySource` directly in tests, so this will cause some PCT issues. Here is what I see in a quick GitHub search:
* https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/41dd8ef6dd561452755641a2a3c76e69e9b2d0cb/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/CredentialWrapperStepTest.java#L118
* https://github.com/jenkinsci/credentials-binding-plugin/blob/dc9d38cb254d73797414717cdb6049150b206f87/src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBindingTest.java#L100
* https://github.com/jenkinsci/compuware-common-configuration-plugin/blob/4a9f696e77a24f1b688d120c7738a55d62d5bd27/src/test/java/com/compuware/jenkins/common/configuration/HostConnectionTest.java#L282

jtnord:
> @jtnord Maybe you already know, but just in case, a few plugins use `FileOnMasterKeyStoreSource` directly in tests, so this will cause some PCT issues. Here is what I see in a quick GitHub search:

****** GitHub search....

image

dwnusbaum:
@jtnord The PR title has a typo, I edited my comment once I realized, see https://github.com/search?type=code&q=+owner%3Ajenkinsci+FileOnMasterKeyStoreSource

jtnord:
> @jtnord The PR title has a typo, I edited my comment once I realized, see https://github.com/search?type=code&q=+owner%3Ajenkinsci+FileOnMasterKeyStoreSource

🤦   thanks!

olamy (approved):

pom.xml
jtnord:
```suggestion
          <compatibleSinceVersion>1354</compatibleSinceVersion>
```

basil:
Breaks PCT in `credentials-binding` with `java.lang.NoClassDefFoundError`: `com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl$FileOnMasterKeyStoreSource`. The class is still used in `CertificateMultiBindingTest`.

basil:
@jtnord

dwnusbaum:
BOM update just needs to be coordinated with https://github.com/jenkinsci/credentials-binding-plugin/pull/310 and https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/723.

basil:
> BOM update just needs to be coordinated with [jenkinsci/credentials-binding-plugin#310](https://github.com/jenkinsci/credentials-binding-plugin/pull/310) and [jenkinsci/pipeline-model-definition-plugin#723](https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/723).

@dwnusbaum No, jenkinsci/credentials-binding-plugin#310 is incomplete—that PR removed one usage of `FileOnMasterKeyStoreSource`, but there is still another usage a few dozen lines down in the same file.

PR #557, opened 2024-09-09
Adding credentials should not require a check on CredentialsStore#isDomainsModifiable

Dohbedoh:
Fix a regression caused by https://github.com/jenkinsci/credentials-plugin/pull/481. Since this PR, store implementation that define methods to manage credentials but not to manage domains cannot use the `Add` button anymore... This is the case for CloudBees implementation for Cyberark / Hashicorp Vault for example. 

It fails with a 400 and the "Domain is read-only" error on screen. Before this PR, the method was adding credentials properly and send return 400. Since this PR, it is returning early with the error.

Trying to make sense of this [condition](https://github.com/jenkinsci/credentials-plugin/blob/1371.vfee6b_095f0a_3/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java#L600-L605), I think it is is wrong. Per the javadoc for [CredentialsStore#isDomainsModifiable()](https://github.com/jenkinsci/credentials-plugin/blob/0d8f3bb8a984b5a4a7cfbce18437d4e5dfcaee31/src/main/java/com/cloudbees/plugins/credentials/CredentialsStore.java#L261-L277), it return if the store supports making changes to the list of domains or whether it only supports a fixed set of domains (which may only be one domain). This is about altering Domains. This method does not update or create a domain in any way, it simply add credentials to a selected domain. If the domain selected does not exist it fails with `Store does not have selected domain`. Also if the domain is not modifiable, you would not be able to select it in the dropdown:\

credentials-domain-add-credentials-broken

So per my understanding, this condition should be removed.

### Submitter checklist


rsandell:
Well this check has been there since before version 2.0 (2016) so what has changed now to cause problems? The check looks very explicit to me.

dwnusbaum:
@rsandell FWIW, separately (BEE-51772), Allan claimed that while present, the check in the old code was not functional, and #481 changed that behavior:

> Before that change, the 400 was already happening but silent. Now the credentials is not created anymore..

Dohbedoh:
> Well this check has been there since before version 2.0 (2016) so what has changed now to cause problems? The check looks very explicit to me.

Since the change from , we return early within the condition: 

* https://github.com/jenkinsci/credentials-plugin/blob/1371.vfee6b_095f0a_3/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java#L602-L604. 

Before that change, the HTTP 400 response was generated but we did not return so the rest of the method got executed as well:

> And it looks kinda correct to me that if a domain isn't modifiable you shouldn't be able to add any credentials to it.

Seems to me that this condition was not effective all along..

> And it looks kinda correct to me that if a domain isn't modifiable you shouldn't be able to add any credentials to it. For example you can't add an object to an unmodifiable list.

My understanding of the javadoc from this method and the usage of that method is that when a CredentialsStore is "domain-unmodifiable", you simply cannot create/delete/modify domains. But you can manage credentials within the read only domain(s). The workaround for that regression for example, is to add credentials from **Manage Jenkins > Credentials** for root credentials or Folder's **Credentials** page for folder credentials.

car-roll:
I've been going back and forth on this. At first my "feeling" was that if a domain is used to categorize credentials, and we are setting that domain to unmodifiable, my expectation is that I should not be adding additional credentials to that domain. But like Alan said, the javadoc says something entirely different:
```Identifies whether this CredentialsStore supports making changes to the **list of domains** or whether it only supports **a fixed set of domains** (which may only be one domain).```

It may have been misapplication of this particular check? I see there is a check if you can [add a credential to a store](https://github.com/jenkinsci/credentials-plugin/blob/0d8f3bb8a984b5a4a7cfbce18437d4e5dfcaee31/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java#L424), but really nothing else for domain.

car-roll (approved):

jeromepochat:
This PR is currently empty of change. Is that expected? The initial changes  https://github.com/jenkinsci/credentials-plugin/pull/557/commits/8dc8ff612fd4b6f6a725473d4ca5b1ac76366d73 are reverted by merge commit https://github.com/jenkinsci/credentials-plugin/pull/557/commits/b933fee17a443981d2441b60f7da7efd57fc6d3a

jeromepochat (approved):

olamy (approved):

PR #558, opened 2024-09-12
In FIPS environment password must be at least 14 characters

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
@@ -34,7 +34,7 @@ import hudson.util.FormValidation;
 import hudson.util.Secret;
 
 import jenkins.security.FIPS140;
-import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang.StringUtils;
 import org.jenkinsci.Symbol;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
--- a/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
@@ -39,6 +39,7 @@ import org.jenkinsci.Symbol;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
+import org.kohsuke.stapler.interceptor.RequirePOST;
 
 import java.util.Objects;
 
@@ -138,6 +139,7 @@ public class UsernamePasswordCredentialsImpl extends BaseStandardCredentials imp
             return "symbol-id-card";
         }
 
+        @RequirePOST
         public FormValidation doCheckPassword(@QueryParameter String password) {
             if(FIPS140.useCompliantAlgorithms() && StringUtils.length(password) < 14) {
                 return FormValidation.error(Messages.passwordTooShortFIPS());
--- a/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
@@ -30,6 +30,7 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import hudson.Extension;
 import hudson.Util;
+import hudson.model.Descriptor;
 import hudson.util.FormValidation;
 import hudson.util.Secret;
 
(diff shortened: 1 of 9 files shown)


src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
```diff
@@ -76,6 +83,9 @@ public UsernamePasswordCredentialsImpl(@CheckForNull CredentialsScope scope,
                                            @CheckForNull String username, @CheckForNull String password) {
         super(scope, id, description);
         this.username = Util.fixNull(username);
+        if(FIPS140.useCompliantAlgorithms() && StringUtils.length(password) < 14) {
+            throw new IllegalArgumentException(Messages.passwordTooShortFIPS());
```
jtnord, line 87:
```suggestion
            throw new hudson.model.Descriptor.FormException(Messages.passwordTooShortFIPS(), "password");
```
?

src/test/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImplFIPSTest.java
```diff
@@ -0,0 +1,90 @@
...
+import static org.hamcrest.Matchers.emptyString;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
+import static org.junit.Assert.assertThrows;
+
+public class UsernamePasswordCredentialsImplFIPSTest {
+
+    @Rule public RealJenkinsRule rule = new RealJenkinsRule().javaOptions("-Djenkins.security.FIPS140.COMPLIANCE=true")
```
jtnord, line 31:
```suggestion
    @Rule public RealJenkinsRule rule = new RealJenkinsRule().javaOptions("-Djenkins.security.FIPS140.COMPLIANCE=true", "-Xmx=512M")
```
otherwise this test may well reserve a lot more ram than it actually needs.

src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java
olamy: in the constructor? Why not but I find it weird in a constructor to have this exception which looks more related to form (web action) but maybe I'm biased by the class name and it's implementing `HttpResponse`

jtnord: yup, it means the user gets a friendlier error page that tells them what they did that went wrong (ish) [JENKINS-73771](https://issues.jenkins.io/browse/JENKINS-73771)

olamy: this will need to note the PR as breaking, as such exception is breaking backward compat.

jtnord: > this will need to note the PR as breaking, as such exception is breaking backward compat.

Add a `compatable since` to the `maven-hpi` configuration so it gets called out in the plugin manager then?

olamy: done with a3924cecba36fe56f4f708a780a21d07e25c7fb3 
should be ok but not really sure about the random number to use version

src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsSnapshotTaker.java
```diff
@@ -25,8 +26,13 @@ public StandardUsernamePasswordCredentials snapshot(StandardUsernamePasswordCred
...
-        UsernamePasswordCredentialsImpl snapshot = new UsernamePasswordCredentialsImpl(credentials.getScope(), credentials.getId(), credentials.getDescription(), credentials.getUsername(), Secret.toString(credentials.getPassword()));
-        snapshot.setUsernameSecret(credentials.isUsernameSecret());
-        return snapshot;
+        try {
+            UsernamePasswordCredentialsImpl snapshot =
+                    new UsernamePasswordCredentialsImpl(credentials.getScope(), credentials.getId(), credentials.getDescription(), credentials.getUsername(), Secret.toString(credentials.getPassword()));
+            snapshot.setUsernameSecret(credentials.isUsernameSecret());
+            return snapshot;
+        } catch (Descriptor.FormException e) {
+            throw new RuntimeException(e);
```
jtnord, line 35: weired.  I was expecting the method to throw be able to throw an `IOException` as many credentials can (and are) documented to do that.  [doc](https://github.com/jenkinsci/credentials-plugin/blob/master/docs/implementation.adoc#:~:text=You%20should%20note%20that%20we%20have%20used%20Secret%20to%20define%20the%20secret.%20And%20furthermore%2C%20we%20have%20added%20the%20throws%20IOException%2C%20InterruptedException%20to%20the%20getter!%20Again%20this%20is%20so%20that%20a%20new%20CredentialsProvider%20can%20lazily%20fetch%20the%20secret.)

jtnord (approved):

basil:
Breaks `hudson.plugins.emailext.MailAccountFIPSTest#testFormValidationForInsecureAuth`, `hudson.plugins.git.FIPSModeUrlCheckTest#testGitSCMSourceCheck`, and `org.jenkinsci.plugins.kubernetes.credentials.OpenShiftBearerTokenCredentialWithFIPSTest`.

PR #559, opened 2024-09-20
Demonstrate weird error

daniel-beck:
Just to see the build fail


jglick:
Presumably obsolete, whatever this was.

daniel-beck:
Sort of, worked around this with https://github.com/jenkinsci/credentials-plugin/blob/9dd7898fafb223d36975622bab2ac6f617a7243f/src/test/java/com/cloudbees/plugins/credentials/SecretBytesRedactionTest.java#L68.

PR #561, opened 2024-10-02
Update user.adoc - Example of adding a 'secret text' credential and correction of a misspelling.

kovacstim:
image


jglick:
It works: https://github.com/jenkinsci/jenkins/blob/c64384a8671b03b0541342b678d414ae51e308be/core/src/main/java/hudson/util/Secret.java#L242-L255

PR #580, opened 2024-12-06
[JENKINS-74964] Display error message when adding invalid certificate credentials

Priya-CB:
[JENKINS-74964](https://issues.jenkins.io/browse/JENKINS-74964)

- The "Add Credentials" popup now correctly displays an error message when users attempt to add invalid certificate credentials. Previously, the validation errors were shown on the form but no error message appeared upon clicking "Add," leading to a silent rejection of the credentials. 

### Testing done

Automated Unit Testing

Implemented unit tests to verify that the error messages are returned in the response on form submit 

Manual Testing

1. Adding certificate credentials with invalid inputs (e.g., empty PEM-encoded certificates or password < 14 characters in FIPS mode).
2. Verified that clicking "Add" prevents invalid credentials from being stored and displays an appropriate error message to the user.

Popup screen:

Screenshot 2024-12-06 at 9 19 12 PM

Screenshot 2024-12-06 at 9 14 04 PM

Screenshot 2024-12-06 at 9 23 39 PM

Screenshot 2024-12-06 at 8 19 14 PM

### Submitter checklist


src/main/resources/lib/credentials/select/select.js
```diff
@@ -123,6 +123,7 @@ window.credentials.addSubmit = function (_) {
             .catch((e) => {
                 // notificationBar.show(...) with logging ID could be handy here?
                 console.error("Could not add credentials:", e);
+                window.notificationBar.show("Credentials creation failed.", window.notificationBar["ERROR"]);
```
Priya-CB, line 126: The current implementation of the "Add Credentials" popup does not handle HTML responses ("angry Jenkins"), causing the catch block to execute. Added notificationBar to inform the user about the failure(unexpected cases) in credential creation.

Priya-CB, line 126: "Add Credentials" popup does not handle HTML responses ("angry Jenkins"), causing the catch block to execute. Added notificationBar to inform the user about the failure(unexpected cases) in credential creation.

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
```diff
@@ -129,17 +129,22 @@ public class CertificateCredentialsImpl extends BaseStandardCredentials implemen
...
+        }
         this.password = Secret.fromString(password);
         this.keyStoreSource = keyStoreSource;
         // ensure the keySore is valid
         // we check here as otherwise it will lead to hard to diagnose errors when used
         try {
             keyStoreSource.toKeyStore(toCharArray(this.password));
         } catch (GeneralSecurityException | IOException e) {
-            throw new IllegalArgumentException("KeyStore is not valid.", e);
+            LOGGER.log(Level.WARNING, Messages.CertificateCredentialsImpl_InvalidKeystore(), e);
```
jtnord, line 145: please keep logs unlocalized.
```suggestion
            LOGGER.log(Level.WARNING, "Failed to create CertificateCredentials", e);
```

src/test/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImplFIPSTest.java
```diff
@@ -0,0 +1,80 @@
...
+import com.cloudbees.plugins.credentials.CredentialsScope;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThrows;
+
+public class CertificateCredentialsImplFIPSTest {
+
+    @Rule
+    public RealJenkinsRule rule = new RealJenkinsRule().javaOptions("-Djenkins.security.FIPS140.COMPLIANCE=true");
```
jtnord, line 27: please use [`RealJenkinsRile.withFipsEnabled`](https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/RealJenkinsRule.html#withFIPSEnabled()) and always specify some memory options so this does not use default ergonomics and grab a huge portion of available memory.

src/test/resources/com/cloudbees/plugins/credentials/impl/validCerts.pem
```diff
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
```
jtnord, line 1: is there something wrong with the [existing](https://github.com/jenkinsci/credentials-plugin/tree/master/src/test/resources/com/cloudbees/plugins/credentials/impl) PEM files that means you can not use them?

src/test/resources/com/cloudbees/plugins/credentials/impl/validKey.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
jtnord, line 1: whilst the current [PEK key](https://github.com/jenkinsci/credentials-plugin/tree/master/src/test/resources/com/cloudbees/plugins/credentials/impl) is encrypted, would it not be better to create a file with the same key without encyrption in the same place?

src/test/resources/com/cloudbees/plugins/credentials/impl/validCerts.pem
Priya-CB: The existing PEM files cannot be used in FIPS mode because the encryption algorithm used is not FIPS compliant.
Error:
org.bouncycastle.crypto.fips.FipsUnapprovedOperationError: salt must be at least 128 bits

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
```diff
@@ -129,17 +129,22 @@ public class CertificateCredentialsImpl extends BaseStandardCredentials implemen
     public CertificateCredentialsImpl(@CheckForNull CredentialsScope scope,
                                       @CheckForNull String id, @CheckForNull String description,
                                       @CheckForNull String password,
-                                      @NonNull KeyStoreSource keyStoreSource) {
+                                      @NonNull KeyStoreSource keyStoreSource) throws Descriptor.FormException {
```
PereBueno, line 132: This signature change in a public method might break some plugins --> https://github.com/search?q=org%3Ajenkinsci+%22new+CertificateCredentialsImpl%22&type=code

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
Priya-CB: My bad, updated it

src/test/resources/com/cloudbees/plugins/credentials/impl/validKey.pem
Priya-CB: I don't have the passphrase for the existing key, to create an unencrypted version. Please correct me if I misunderstood your comment.

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
Priya-CB: Yes, the change is binary-compatible but source-incompatible. The [Descriptor#newInstanceImpl](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/Descriptor.java#L623) throws a LinkageError if an IllegalArgumentException is thrown from the constructor. Using exceptions like Descriptor.FormException ensures that errors are rendered with a helpful error message instead of the angry Jenkins page.

Please let me know if there is a better approach.

src/test/resources/com/cloudbees/plugins/credentials/impl/validKey.pem
jtnord: it is in the source code (tests) where it is used -> https://github.com/jenkinsci/credentials-plugin/blob/1e306e8f6c54f8eeef973b3387b8c223bd60ae1c/src/test/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImplTest.java#L297-L309

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
```diff
@@ -608,8 +616,30 @@ public JSONObject doAddCredentials(StaplerRequest2 req, StaplerResponse2 rsp) th
...
+                 *
+                 * This approach is taken to maintain backward compatibility, as throwing a FormException directly
+                 * from the constructor would result in a source-incompatible change, potentially breaking dependent plugins.
+                 *
+                 * Here, known exceptions are caught specifically to provide meaningful error response.
+                 */
+                Throwable rootCause = ExceptionUtils.getRootCause(e);
+                if (rootCause instanceof IOException || rootCause instanceof IllegalArgumentException
+                    || rootCause instanceof GeneralSecurityException) {
+                    LOGGER.log(Level.WARNING, "Failed to create Credentials", e);
```
pankajy-dev, line 637: Should we have a log level of type error?

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
```diff
@@ -608,8 +616,30 @@ public JSONObject doAddCredentials(StaplerRequest2 req, StaplerResponse2 rsp) th
...
-            boolean credentialsWereAdded = store.addCredentials(wrapper.getDomain(), credentials);
+            boolean credentialsWereAdded;
+            try {
+                Credentials credentials = Descriptor.bindJSON(req, Credentials.class,
+                                                              data.getJSONObject("credentials"));
+                credentialsWereAdded = store.addCredentials(wrapper.getDomain(), credentials);
+            } catch (LinkageError e) {
+                /*
+                 * Descriptor#newInstanceImpl throws a LinkageError if the constructor throws any exception other
+                 * than HTTP response exceptions such as Descriptor.FormException.
```
fcojfernandez, line 627: AFAIK `FormException ` is not a `RuntimeException`, and only `RuntimeException are not causing a LinkageError: https://github.com/jenkinsci/jenkins/blob/171097618b5b2b70ec3142e3efc80b4f2aff9601/core/src/main/java/hudson/model/Descriptor.java#L623 so
```suggestion
                 * Descriptor#newInstanceImpl throws a LinkageError if the DataBoundConstructor or any DataBoundSetter throw any exception other
                 * than RuntimeException implementing HttpResponse.
```

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
fcojfernandez: Next `Level` would be `SEVERE` that sounds too strong to me. I'd leave the`WARNING` as the instance (or simply the page calling the credentials popup) won't crash. It's just the credentials not created

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
```diff
@@ -608,8 +616,30 @@ public JSONObject doAddCredentials(StaplerRequest2 req, StaplerResponse2 rsp) th
...
+                 *
+                 * This approach is taken to maintain backward compatibility, as throwing a FormException directly
+                 * from the constructor would result in a source-incompatible change, potentially breaking dependent plugins.
+                 *
+                 * Here, known exceptions are caught specifically to provide meaningful error response.
+                 */
+                Throwable rootCause = ExceptionUtils.getRootCause(e);
+                if (rootCause instanceof IOException || rootCause instanceof IllegalArgumentException
+                    || rootCause instanceof GeneralSecurityException) {
```
fcojfernandez, line 636: Why the 3 exceptions? AFAICT only `IllegalArgumentException` will be the root exception, right?

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
jmdesprez: I agree to keep the `WARNING` level here (nice suggestion anyway)

jmdesprez (commented):
I agree with previous reviews/comments

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
Priya-CB: While FormException itself is not a RuntimeException, it is wrapped into HttpResponses, which is a RuntimeException,
https://github.com/jenkinsci/stapler/blob/master/core/src/main/java/org/kohsuke/stapler/RequestImpl.java#L691

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
Priya-CB: > Why the 3 exceptions? AFAICT only IllegalArgumentException will be the root exception, right?

The reason for handling IOException and GeneralSecurityException along side IllegalArgumentException is due to how exceptions are wrapped and propagated. In this case, the constructor itself catches IOException and GeneralSecurityException and rethrows them as IllegalArgumentException. However, when these exceptions bubble up to this catch block, they are wrapped at multiple levels, ultimately resulting in a LinkageError.
When we retrieve the root cause, it points back to the original IOException or GeneralSecurityException rather than the rethrown IllegalArgumentException.
https://github.com/jenkinsci/credentials-plugin/blob/master/src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java#L703
https://github.com/jenkinsci/credentials-plugin/blob/master/src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java#L142

> Also, is it possible in this plugin to see the LinkageError with a different root cause than the IllegalArgumentException? Why is the if needed?

Regarding LinkageError, the check for different root causes is a safeguard. Although IllegalArgumentException is the expected root cause in most scenarios, this ensures that other potential but unexpected root causes (like related to class loading or dependencies) are not overlooked and are handled appropriately.

jmdesprez (approved):

pankajy-dev (approved):

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
fcojfernandez: no, that's incorrect. As for https://github.com/jenkinsci/jenkins/blob/171097618b5b2b70ec3142e3efc80b4f2aff9601/core/src/main/java/hudson/model/Descriptor.java#L622-L627 the Linkage error only can happen if the caught exception is RuntimeException AND HttpResponse. FormException is not in the catch clause, so it's not caught at all but directly thrown as stated in the method signature. The only way of getting a LinkageError is that the error is a NoSuchMethodException, an InstantiationException, an IllegalAccessException,  an InvocationTargetException, or a RuntimeException that is not implementing HttpResponse

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
fcojfernandez: > In this case, the constructor itself catches IOException and GeneralSecurityException and rethrows them as IllegalArgumentException

but for those cases, the root cause of the LinkageError would be the IllegalArgumentException and not the IOException or the GeneralSecurityException. Or am I missing something?

fcojfernandez: > > Also, is it possible in this plugin to see the LinkageError with a different root cause than the IllegalArgumentException? Why is the if needed?
> 
> Regarding LinkageError, the check for different root causes is a safeguard. Although IllegalArgumentException is the expected root cause in most scenarios, this ensures that other potential but unexpected root causes (like related to class loading or dependencies) are not overlooked and are handled appropriately.

fair enough

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
Priya-CB: You are correct that FormException is not explicitly in the catch clause of Descriptor#newInstanceImpl. However, my point was that FormException, when thrown, is wrapped into HttpResponses and HttpResponses is a RuntimeException.
Because of this wrapping, it ultimately satisfies the condition of being a RuntimeException and implementing HttpResponse

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
Priya-CB: > > In this case, the constructor itself catches IOException and GeneralSecurityException and rethrows them as IllegalArgumentException
> 
> but for those cases, the root cause of the LinkageError would be the IllegalArgumentException and not the IOException or the GeneralSecurityException. Or am I missing something?

The constructor catches IOException and GeneralSecurityException and rethrows them as an IllegalArgumentException. This means that IllegalArgumentException becomes the direct exception thrown, and the original exceptions (IOException or GeneralSecurityException) are set as its cause.

When LinkageError is thrown, it is wrapping the IllegalArgumentException. However, ExceptionUtils#getRootCause traverse the entire chain of exceptions and stop at the deepest cause.

In this case, the deepest cause (root cause) would still be the original IOException or GeneralSecurityException, not the rethrown IllegalArgumentException.

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
Priya-CB: Updated the comment

src/test/resources/com/cloudbees/plugins/credentials/impl/validKey.pem
Priya-CB: Added the FIPS complaint encrypted key and certificate

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
Priya-CB: Removed the incompatible changes and handled the LinkageError in CredentialsSelectHelper#doAddCredentials

fcojfernandez (approved):

PR #581, opened 2024-12-07
Handle password with empty string for certificate auth

timja:
image

### Testing done

Uploaded a cert generated inside Azure Key Vault

Tested with https://github.com/jenkinsci/azure-credentials-plugin/pull/274
and checked end 2 end and I can provision VMs with https://plugins.jenkins.io/azure-vm-agents using certificate auth

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
@@ -147,12 +147,10 @@ public class CertificateCredentialsImpl extends BaseStandardCredentials implemen
      * Helper to convert a {@link Secret} password into a {@code char[]}
      *
      * @param password the password.
-     * @return a {@code char[]} containing the password or {@code null}
+     * @return a {@code char[]} containing the password
      */
-    @CheckForNull
     private static char[] toCharArray(@NonNull Secret password) {
-        String plainText = Util.fixEmpty(password.getPlainText());
-        return plainText == null ? null : plainText.toCharArray();
+        return password.getPlainText().toCharArray();
     }
 
     /**
@@ -248,7 +246,7 @@ public class CertificateCredentialsImpl extends BaseStandardCredentials implemen
                 return FormValidation.error(Messages.CertificateCredentialsImpl_ShortPasswordFIPS());
             }
             if (pw.isEmpty()) {
-                return FormValidation.warning(Messages.CertificateCredentialsImpl_NoPassword());
+                return FormValidation.ok(Messages.CertificateCredentialsImpl_NoPassword());
             }
             if (pw.length() < 14) {
                 return FormValidation.warning(Messages.CertificateCredentialsImpl_ShortPassword());
@@ -624,9 +622,7 @@ public class CertificateCredentialsImpl extends BaseStandardCredentials implemen
                 } catch (KeyStoreException | CertificateException | NoSuchAlgorithmException | IOException e) {
                     return FormValidation.warning(e, Messages.CertificateCredentialsImpl_LoadKeystoreFailed());
                 } finally {
-                    if (passwordChars != null) {
-                        Arrays.fill(passwordChars, ' ');
-                    }
+                    Arrays.fill(passwordChars, ' ');
                 }
             }
 
@@ -739,6 +735,9 @@ public class CertificateCredentialsImpl extends BaseStandardCredentials implemen
                     List<PEMEncodable> pemEncodables = PEMEncodable.decodeAll(pemCerts, null);
                     long count = pemEncodables.stream().map(PEMEncodable::toCertificate).filter(Objects::nonNull).count();
                     if (count < 1) {
+                        if (Util.fixEmpty(value) == null) {
+                            return FormValidation.ok();
+                        }
                         return FormValidation.error(Messages.CertificateCredentialsImpl_PEMNoCertificates());
                     }
                     // ensure only certs are provided.
@@ -771,6 +770,9 @@ public class CertificateCredentialsImpl extends BaseStandardCredentials implemen
                     List<PEMEncodable> pemEncodables = PEMEncodable.decodeAll(key, toCharArray(Secret.fromString(password)));
                     long count = pemEncodables.stream().map(PEMEncodable::toPrivateKey).filter(Objects::nonNull).count();
                     if (count == 0) {
+                        if (Util.fixEmpty(value) == null) {
+                            return FormValidation.ok();
+                        }
                         return FormValidation.error(Messages.CertificateCredentialsImpl_PEMNoKeys());
                     }
                     if (count > 1) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
```diff
@@ -147,12 +147,10 @@ public CertificateCredentialsImpl(@CheckForNull CredentialsScope scope,
...
      *
      * @param password the password.
-     * @return a {@code char[]} containing the password or {@code null}
+     * @return a {@code char[]} containing the password
      */
-    @CheckForNull
     private static char[] toCharArray(@NonNull Secret password) {
-        String plainText = Util.fixEmpty(password.getPlainText());
-        return plainText == null ? null : plainText.toCharArray();
+        return password.getPlainText().toCharArray();
```
timja, line 153: Empty string is used at least commonly in Azure if not elsewhere for PFX files.

Shouldn't modify what the user has passed here

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
```diff
@@ -248,7 +246,7 @@ public FormValidation doCheckPassword(@QueryParameter String value) {
                 return FormValidation.error(Messages.CertificateCredentialsImpl_ShortPasswordFIPS());
             }
             if (pw.isEmpty()) {
-                return FormValidation.warning(Messages.CertificateCredentialsImpl_NoPassword());
+                return FormValidation.ok(Messages.CertificateCredentialsImpl_NoPassword());
```
timja, line 249: 1. warnings/errors shouldn't be shown when you open a form
2. this is a valid usecase but its still worth making people aware so I left it as ok

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
```diff
@@ -624,9 +622,7 @@ protected static FormValidation validateCertificateKeystore(byte[] keystoreBytes
                 } catch (KeyStoreException | CertificateException | NoSuchAlgorithmException | IOException e) {
                     return FormValidation.warning(e, Messages.CertificateCredentialsImpl_LoadKeystoreFailed());
                 } finally {
-                    if (passwordChars != null) {
```
timja, line 627: no longer possible to be null here

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
```diff
@@ -739,6 +735,9 @@ public FormValidation doCheckCertChain(@QueryParameter String value) {
                     List<PEMEncodable> pemEncodables = PEMEncodable.decodeAll(pemCerts, null);
                     long count = pemEncodables.stream().map(PEMEncodable::toCertificate).filter(Objects::nonNull).count();
                     if (count < 1) {
+                        if (Util.fixEmpty(value) == null) {
```
timja, line 738: warnings/errors shouldn't be shown when you open a form

The form has a clear empty state so we don't need the same content as we did with the password for the pfx:

image

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
```diff
@@ -771,6 +770,9 @@ public FormValidation doCheckPrivateKey(@QueryParameter String value,
                     List<PEMEncodable> pemEncodables = PEMEncodable.decodeAll(key, toCharArray(Secret.fromString(password)));
                     long count = pemEncodables.stream().map(PEMEncodable::toPrivateKey).filter(Objects::nonNull).count();
                     if (count == 0) {
+                        if (Util.fixEmpty(value) == null) {
```
timja, line 773: warnings/errors shouldn't be shown when you open a form

The form has a clear empty state so we don't need the same content as we did with the password for the pfx:

image

src/main/resources/com/cloudbees/plugins/credentials/impl/Messages.properties
```diff
@@ -24,8 +24,8 @@
 UsernamePasswordCredentialsImpl.DisplayName=Username with password
 CertificateCredentialsImpl.DisplayName=Certificate
 CertificateCredentialsImpl.EmptyKeystore=Empty keystore
-CertificateCredentialsImpl.LoadKeyFailed=Could retrieve key "{0}"
-CertificateCredentialsImpl.LoadKeyFailedQueryEmptyPassword=Could retrieve key "{0}". You may need to provide a password
```
timja, line 28: these error messages made no sense before

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
```diff
@@ -147,12 +147,10 @@ public CertificateCredentialsImpl(@CheckForNull CredentialsScope scope,
      * Helper to convert a {@link Secret} password into a {@code char[]}
      *
      * @param password the password.
-     * @return a {@code char[]} containing the password or {@code null}
+     * @return a {@code char[]} containing the password
      */
-    @CheckForNull
```
jtnord, line 152: There is a difference between no password and an empty password.

It seems like Azure is using the empty password rather than no password (why I have no idea).

https://stackoverflow.com/a/53523999 appears to confirm this.

jtnord (commented):
I think this fixes the issue on azure, but breaks for others where there is no password at all (rather than an empty password)

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
timja: Not sure looks like other systems try both empty string and null

jtnord:
Thanks for the investigation and fix

src/main/java/com/cloudbees/plugins/credentials/impl/CertificateCredentialsImpl.java
jtnord: I would agree if Jenkins did not allow you to hit "Save" or Apply when the form was in an invalid state or otherwise called out mandatory parameters from optional ones, but it does neither.

Hitting save leads to an angry Jenkins which is IMO a worse UX that the error.  

```
2024-12-10 11:23:30.372+0000 [id=41]    WARNING h.i.i.InstallUncaughtExceptionHandler#handleException: Caught unhandled exception with ID a655cb61-a72d-4d07-a85b-97a161a42acd
java.io.IOException: expected one key but got 0
        at PluginClassLoader for credentials//com.cloudbees.plugins.credentials.impl.CertificateCredentialsImpl$PEMEntryKeyStoreSource.toKeyStore(CertificateCredentialsImpl.java:703)
        at PluginClassLoader for credentials//com.cloudbees.plugins.credentials.impl.CertificateCredentialsImpl$PEMEntryKeyStoreSource.toKeyStore(CertificateCredentialsImpl.java:691)
        at PluginClassLoader for credentials//com.cloudbees.plugins.credentials.impl.CertificateCredentialsImpl.<init>(CertificateCredentialsImpl.java:140)
Caused: java.lang.IllegalArgumentException: KeyStore is not valid.
```

PR #602, opened 2025-03-18
improve dialog to add credentials

mawinter69:
apply jenkins style to the domain select

set minwidth of dialog via min function so it doesn't get too wide on large screens

Before:
Dialog was always 75% of screen width adding a large blank area to the right in the dialog on large screens
inputs are not right aligned
Domain select with old styling
image

After:
Large screen 
image

Small Screen:
all inputs are properly right aligned
image

### Submitter checklist


MarkEWaite (approved):
Thanks!

PR #627, opened 2025-06-03
Remove action.jelly

janfaracik:
_As far as I can tell_, these files aren't used/visible in the interface - seeing as a lot of the pages are 'one-column' I'm not sure where'd they'd appear.

### Testing done

* Interface still functional
* Breadcrumbs still work

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

PR #658, opened 2025-11-13
[JENKINS-76290] Fix ViewCredentialsAction appearing for items without credentials stores

apuig:
[JENKINS-76290 - Credentials Link Appears for Regular Jobs](https://issues.jenkins.io/browse/JENKINS-76290)

`ViewCredentialsAction.isVisible()` was returning true for any item where a `CredentialsProvider` had descriptors, without verifying that provider actually returned a store for that specific item.

* `hasCredentialsDescriptors()`  is context-independent:  "can provide credentials in general?". **uses credentials** 
* `getStore(context)`  is context-dependent: "provides credentials for THIS item?". **manage credentials**

This caused the credentials link to appear in the sidebar for non-folder items like, jobs, or other TopLevelItems that don't have credentials stores. Clicking the link resulted in a 403 Forbidden error.

The fix adds a check to ensure the provider returns a non-null store for the context before considering the action visible.

Probably regression introduced in jenkinsci/credentials-plugin#627 which removed `action.jelly` with explicit permission check.

Example: `FolderCredentialsProvider.hasCredentialsDescriptors()` returns `true` for any item, but `getStore(item)` only returns non-null for folders.

### Testing done
* Added a new test and manually verified on a FreeStyleProject

### Submitter checklist

### DIFF
--- a/src/test/java/com/cloudbees/plugins/credentials/ViewCredentialsActionTest.java
+++ b/src/test/java/com/cloudbees/plugins/credentials/ViewCredentialsActionTest.java
@@ -2,6 +2,7 @@ package com.cloudbees.plugins.credentials;
 
 import static com.cloudbees.plugins.credentials.XmlMatchers.isSimilarToIgnoringPrivateAttrs;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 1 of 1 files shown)


apuig:
Hi @jtnord, thank you for your review.
Is there anything I can do to help move it forward? If you think it would be helpful, perhaps you could suggest an additional reviewer. 

Thanks again for your support!

PR #984, opened 2025-12-19
Use card layout for credentials tables

timja:
Goal of this is to reduce duplication of information in e.g.:
image

Whilst making it easier to read and understand.
Also should stop credentials with longer names of descriptions from overflowing and breaking out of the page.

<details>
<summary>Root action before</summary>

image

</details>

<details>
<summary>Root action after</summary>

image

</details>

The above highlights that the display names are not the best for all credential types. I think its optimised for the credentials picker dropdown, but not great for the credentials list, I'd like to look at this if I have time but not for this PR

<details>
<summary>Domain</summary>

image

</details>

<details>
<summary>Empty state domain modifiable</summary>

image

</details>

<details>
<summary>Empty state domain not modifiable</summary>

image

</details>

<details>
<summary>Empty state root action</summary>

image

</details>

### Testing done

Added a few credentials of differing types.
Added ones without description
Tested with read only stores (Azure key vault)

Ath passed as at https://github.com/jenkinsci/credentials-plugin/pull/984/commits/b8944f95dbe886d5b0b63f0f17c3854ea4823891 in https://github.com/jenkinsci/acceptance-test-harness/pull/2565

cc @janfaracik 

### Submitter checklist

### DIFF
--- a/src/test/java/com/cloudbees/plugins/credentials/ViewCredentialsActionTest.java
+++ b/src/test/java/com/cloudbees/plugins/credentials/ViewCredentialsActionTest.java
@@ -2,8 +2,8 @@ package com.cloudbees.plugins.credentials;
 
 import static com.cloudbees.plugins.credentials.XmlMatchers.isSimilarToIgnoringPrivateAttrs;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.junit.Assert.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 


janfaracik:
Looking good! 

Just a few 8am ideas:

* Any thoughts on having the list styled like the dashboard prototype? (`experimental-dashboard-refine`)
  * My current thinking is is that items that can be moved around are separated, e.g. a vertical gap in between them, items that cannot be moved (e.g. jobs, credentials, New Item page) are joined
* Any thoughts on more use of dialogs? E.g. the add credentials page, delete, move
  * And if they are dialogs, do we even need the pages for the credentials? Is it common to deep link to credentials?
* Any thoughts on showing usages on the item itself? (e.g. used N times)

... could I get commit permissions to your fork 🙏

timja:
> Looking good!
> 
> Just a few 8am ideas:
> 
> * Any thoughts on having the list styled like the dashboard prototype? (`experimental-dashboard-refine`)
>   
>   * My current thinking is is that items that can be moved around are separated, e.g. a vertical gap in between them, items that cannot be moved (e.g. jobs, credentials, New Item page) are joined

Sure sounds good

> * Any thoughts on more use of dialogs? E.g. the add credentials page, delete, move

Definitely delete and move could be dialogs, they are very simple pages.
New credentials probably as well.

>   * And if they are dialogs, do we even need the pages for the credentials? Is it common to deep link to credentials?
> * Any thoughts on showing usages on the item itself? (e.g. used N times)

Makes sense, (assuming not expensive to get)

> ... could I get commit permissions to your fork 🙏

Done, go to https://github.com/timja/credentials-plugin/invitations

janfaracik:
Adjusted the cards:

image

There's a whole bunch of fancy focus states that aren't in this (that are in the dashboard concept) - but I'd like to have this be a component in core we can just reuse rather than reinvent the wheel every time.

timja:
@janfaracik thanks for the help!

I've fixed up a few permissions issues where it wasn't quite right for credentials providers that you can't update from directly inside Jenkins (I tested with the Azure Key Vault plugin).

I've also added empty states for when there's no credentials.

Not the biggest fan of the root tables one as I haven't added a call to action due to there being multiple stores and how you can disable them etc I’d like to add this in a follow up

I've created a PR to ATH as well.

Anything else before this is ready do you think?

janfaracik:
> Anything else before this is ready do you think?

Looks good to me 🚀

janfaracik:
image

Can anybody think of a good ionicon to replace this? (https://ionic.io/ionicons)

I can always make one if needed.

timja:
> image
> Can anybody think of a good ionicon to replace this? ([ionic.io/ionicons](https://ionic.io/ionicons))
> 
> I can always make one if needed.

server-outline or business-outline?

FYI there's also icons showing on the action overflow menus that should be updated. (Move etc)

janfaracik:
> server-outline or business-outline?

Ugh I was searching for _tower_ but couldn't find anything - business works well.

Screenshot 2025-12-22 at 09 03 05

> FYI there's also icons showing on the action overflow menus that should be updated. (Move etc)

Aye - will do those in a follow up once we move to modals, hoping to kill the breadcrumb actions...

timja:
Ideally shouldn't have update in the more actions menu and in the main area:
image

(I do wonder if we need the more actions or the update it makes it a bit quicker but the only frequent actions would be new credentials and update a credential

timja:
@jglick possible to take a look at this please?

src/main/resources/lib/credentials/store.jelly
```diff
@@ -0,0 +1,57 @@
...
+                </a>
+            </div>
+            <div class="credentials-card__tags">
+                ${%Domains:}
+                <j:forEach var="domain" items="${store.domains}">
+                    <a href="${storeAction==null?store.getRelativeLinkTo(domain):store.relativeLinkToContext+'credentials/store/'+storeAction.urlName+'/'+domain.url}"
+                       class="jenkins-badge">
+                        <j:choose>
+                            <j:when test="${domain.global}">
+                                ${%global}
```
jglick: nearly illegible for me in Firefox
Screenshot from 2025-12-30 12-40-23

jglick (approved):
A display nit. Otherwise seems OK to me from very brief local testing, but I lack any context for why you are redesigning GUI here. If you plan to work on a lot of stuff like this you should just request write access to the repo.

jtnord:
@timja this removed the model links from the "Stores" section of the overview page so you can no longer quickly add credentials or domains from the overview page, now you need to go to another page and then click introducing an uneeded step?   This to me is a UX regression for power users, and I hope this was unintentional and can be restored?

timja:
> @timja this removed the model links from the "Stores" section of the overview page so you can no longer quickly add credentials or domains from the overview page, now you need to go to another page and then click introducing an uneeded step? This to me is a UX regression for power users, and I hope this was unintentional and can be restored?

New credentials is going to be added this week in https://github.com/jenkinsci/credentials-plugin/pull/992 with a better experience.

Is new domains really needed? Are they used much at all?

jtnord:
> New credentials is going to be added this week in #992 with a better experience.

❤️ 

> Is new domains really needed? Are they used much at all?

I personally rarely use them, and there are none that I see in CloudBees' instances.  But that is possibly as they are currently hard to understand (there is no reason why a github PAT should not be in a domain for github.com - I think its partially the UX to create them is messy and does not guide people to use them, and partially that they do not restrict credentials in them to being used on those domains, but rather restricts requests with a domain to only those credentials).  TBH as they are so rarely used we could possibly change that behaviour to make them actually more useful!

timja:
The new credentials is ready now if you want to take a look. I’m working on ATH and will run it through pct although not expecting anything there

jglick:
IMO the concept of credentials domains was never needed and could be deprecated.

src/main/java/com/cloudbees/plugins/credentials/CredentialsStore.java
```diff
@@ -529,6 +531,21 @@ public String getRelativeLinkTo(Domain domain) {
         return relativeLink + domain.getUrl();
     }
 
+    /**
+     * Returns the icon class name of the {@link #getContext()} of this {@link CredentialsStore}.
+     *
+     * @return the icon class name for the store.
+     * @since TODO
+     */
+    public final String getIconClassName() {
```
jglick: This broke several `CredentialsStore` implementations in CloudBees CI which were implementing `IconSpec` (and thus defining their own `getIconClassName` methods). Is there a recommended migration?

PR #991, opened 2025-12-27
New credentials dialog (basic)

timja:
From a domain:
image

From root action selecting store and domain:
image

Creating the credential from the root action:
image

TODO:
* Allow setting custom class for `l:dialog` to allow styling
  * currently affecting all dialogs which makes the delete one look ugly
* I'd like to add this to the root action page too, complicated by needing to select the store and the domain first
* A radio button with card style like new job or dark theme  for selecting the credential type would be good
* ATH

This could be detached from the cards PR but then would have to reimplement empty state on top

cc @janfaracik

### Testing done

Tested various credentials types.
Adjusted styling to improve design / reduce layout shifting

I expect ATH would need changes

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
@@ -126,29 +126,17 @@ public class CredentialsSelectHelper extends Descriptor<CredentialsSelectHelper>
     }
 
     /**
-     * TODO
-     * @return
+     * @return modifiable store actions for the context provided.
      */
     @Restricted(NoExternalUse.class)
-    public Map<String, List<CredentialsStoreAction.DomainWrapper>> getModifiableStoreActions(ModelObject context, boolean includeUser) {
-//        if (context == null) {
-//            context = CredentialsDescriptor.findContextInPath(ModelObject.class);
-//        }
-
-        System.out.println("");
-        System.out.println("");
-        System.out.println("CONTEXT BELOW");
-        System.out.println(context);
-        System.out.println("");
-        System.out.println("");
-
+    public Map<String, List<CredentialsStoreAction.DomainWrapper>> getModifiableStoreActions(ModelObject context) {
         return StreamSupport.stream(CredentialsProvider.lookupStores(context).spliterator(), false)
                 .filter(s -> s.hasPermission(CredentialsProvider.CREATE))
                 .map(CredentialsStore::getStoreAction)
                 .filter(Objects::nonNull)
                 .collect(Collectors.toMap(
-                        CredentialsStoreAction::getDisplayName,   // "store action name"
-                        store -> new ArrayList<>(store.getDomains().values()),       // "store.domains"
+                        CredentialsStoreAction::getDisplayName,
+                        store -> new ArrayList<>(store.getDomains().values()),
                         (left, right) -> {
                             left.addAll(right);
                             return left;
--- a/src/main/resources/com/cloudbees/plugins/credentials/ViewCredentialsAction/index.jelly
+++ b/src/main/resources/com/cloudbees/plugins/credentials/ViewCredentialsAction/index.jelly
@@ -38,8 +38,7 @@
 
     <l:main-panel>
       <j:set var="selectHelper" value="${app.getDescriptorByName('com.cloudbees.plugins.credentials.CredentialsSelectHelper')}"/>
-      <j:set var="includeUser" value="${attrs.includeUser ? true : false}"/>
-      <j:set var="modifiableStoreActions" value="${selectHelper.getModifiableStoreActions(it, false)}"/>
+      <j:set var="modifiableStoreActions" value="${selectHelper.getModifiableStoreActions(it.context)}"/>
 
       <l:app-bar title="${%Credentials}">
         <l:overflowButton text="${%Add Credentials}"


timja:
There's a WIP for adding it to the root page here: https://github.com/jenkinsci/credentials-plugin/compare/new-credentials-dialog-with-root?expand=1

specifically: https://github.com/jenkinsci/credentials-plugin/commit/aefdd3d4d2c6b91e8be5effe08a48e003755ab52

Not really sure how to do multiple pages or do we just build it up like other pages do.

timja:
Got it working on the root page, pretty rough though:
https://github.com/jenkinsci/credentials-plugin/pull/991/commits/0ccee246e82dda0d50edd8a9fe47c14b022fdc5a

Would be nice to make it a wizard I think rather than nesting but thoughts @janfaracik ?

janfaracik:
image

Any thoughts on something like this? (icons/design would need tweaking).

But basic idea would be to have the user pick the store/domain first, then keep the dialogs consistent between pages.

We should avoid the nesting if we can - would be good to even change the nested `Kind` select to not nest - but not urgent.

timja:
Much nicer

janfaracik:
> Much nicer

Grand - want me to make the change?

timja:
Yes please!

timja:
Let me know if you need a hand, not fully sure how to pass the context through.

Including each dialog on the page would be pretty easy and only showing the right one, but would be nice to have a smarter solution.

janfaracik:
> Let me know if you need a hand, not fully sure how to pass the context through.
> 
> Including each dialog on the page would be pretty easy and only showing the right one, but would be nice to have a smarter solution.

Yeah I'm investigating a wizard approach right now - I can somewhat envisage how I want it to look it's just getting plain JS/Stapler to do that is tricky.

janfaracik:
Check out `janfaracik/new-credentials-wizard` - it's very much broken but the structure is there. I've copied the styling from `jenkins janfaracik/new-item-refinement` for this - seems like it's something we'd want to reuse across core and plugins.

There's an issue with `null is not an object (evaluating 'document.getElementById(id).groupingNode = true')` which I can't quite figure out.

Screenshot 2025-12-29 at 11 28 52

Screenshot 2025-12-29 at 11 29 00

timja:
> There's an issue with `null is not an object (evaluating 'document.getElementById(id).groupingNode = true')` which I can't quite figure out.

Seems the JS is running slightly early, if you run the JS after the error happens it does work... e.g. The expected element exists later

timja:
Think I fixed that few issues I saw:
1. Scope isn't loading
2. 403 error no included crumb when try to create the cred
3. Can't see a way to go back once credential is selected
4. Non modifiable stores are being included currently, (e.g. I can see my non modifiable Azure store)
5. Dialog should be longer I think as next button is off screen on 13" monitor with a few credential types

janfaracik:
> Think I fixed that few issues I saw:
> 
> 1. Scope isn't loading
> 2. 403 error no included crumb when try to create the cred
> 3. Can't see a way to go back once credential is selected
> 4. Non modifiable stores are being included currently, (e.g. I can see my non modifiable Azure store)
> 5. Dialog should be longer I think as next button is off screen on 13" monitor with a few credential types

Am I good to merge `new-credentials-wizard` into this PR and we can fix from there?

timja:
> > Think I fixed that few issues I saw:
> > 
> > 1. Scope isn't loading
> > 2. 403 error no included crumb when try to create the cred
> > 3. Can't see a way to go back once credential is selected
> > 4. Non modifiable stores are being included currently, (e.g. I can see my non modifiable Azure store)
> > 5. Dialog should be longer I think as next button is off screen on 13" monitor with a few credential types
> 
> Am I good to merge `new-credentials-wizard` into this PR and we can fix from there?

yes

timja:
(merged it in)

janfaracik:
Progress

- [ ] Scope isn't loading
- [ ] 403 error no included crumb when try to create the cred
- [x] Can't see a way to go back once credential is selected
- [x] Non modifiable stores are being included currently, (e.g. I can see my non modifiable Azure store)
- [x] Dialog should be longer I think as next button is off screen on 13" monitor with a few credential types

src/main/java/com/cloudbees/plugins/credentials/CredentialsSelectHelper.java
```diff
@@ -121,6 +125,38 @@ public ModelObject resolveContext(Object context) {
...
+//        }
+
+        System.out.println("");
+        System.out.println("");
+        System.out.println("CONTEXT BELOW");
+        System.out.println(context);
+        System.out.println("");
+        System.out.println("");
+
+        return StreamSupport.stream(CredentialsProvider.lookupStores(context).spliterator(), false)
```
timja: user and folder stores are missing here currently

src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/DomainWrapper/dialog2.jelly
```diff
@@ -0,0 +1,45 @@
...
+    <st:setHeader name="X-Wizard-Title" value="Add ${descriptor.displayName}" />
+
+    <l:ajax>
+        <f:form action="${it.store.getRelativeLinkTo(it.domain)}createCredentials" method="POST" name="newCredentials">
+            <div style="margin-top: 0.5rem; display: flex; flex-direction: column; gap: 1.25rem">
+                <div>
+                    <f:rowSet name="credentials">
+                        <j:set var="instance" value="${null}"/>
+                        <f:class-entry descriptor="${descriptor}"/>
+                        <st:include from="${descriptor}" page="${descriptor.configPage}"/>
```
timja: Password isn't redacted in username and password:
image

janfaracik:
I don't have access to commit to this branch 💀 I've pushed a fix for the stores not showing as expected to the previous branch.

timja:
> I don't have access to commit to this branch 💀 I've pushed a fix for the stores not showing as expected to the previous branch.

you should =/

timja:
Oh oops this is an origin branch, I'll re-create it

timja:
now https://github.com/jenkinsci/credentials-plugin/pull/992

PR #1000, opened 2026-01-05
Restore compatibility for `CredentialsStore` implementing `IconSpec`

jglick:
https://github.com/jenkinsci/credentials-plugin/pull/984#discussion_r2662851525 Fixes automated tests but I have not checked it interactively.


timja (approved):
Looks sensible

jglick:
Appears to work in one screen though I am not entirely sure where in the GUI everything is used.

timja:
> Appears to work in one screen though I am not entirely sure where in the GUI everything is used.

Its only used on the ViewCredentialsAction:

image

jglick:
Yeah and there with this patch a custom store does display a custom icon. At least in a functional test which sets up the store somehow. I am not actually sure how to do that manually.

PR #1001, opened 2026-01-06
Deleting unused credentials query language

jglick:
Reverts #55 (except `ListBoxModel` changes with no relation to the rest that I can see). I had noticed in https://github.com/jenkinsci/credentials-plugin/pull/112#discussion_r287355362 that this was unused but https://github.com/jenkinsci/credentials-plugin/pull/112#discussion_r287558723 did not explain any better why #737 was necessary. Perhaps something about lazily loading credentials from an external store? But as per #980 this would best be done by streamlining the process of loading a specific credentials item by id, rather than by creating some hypothetical query translator. I cannot follow what the original plan had been for usage of CQL in some unspecified implementation, but I have searched @jenkinsci and @cloudbees sources for usages of these APIs:

* `CredentialsMatcher.CQL`
* `CredentialsMatchers.describe`
* `CredentialsMatchers.parse`
* `CredentialsMatchers.withProperty`
* `BeanPropertyMatcher`

and found nothing except for usages removed in https://github.com/jenkinsci/bitbucket-branch-source-plugin/pull/940 (released [over a year ago](https://github.com/jenkinsci/bitbucket-branch-source-plugin/releases/tag/933.2.0)).

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java
@@ -1214,10 +1214,8 @@ public abstract class CredentialsProvider extends Descriptor<CredentialsProvider
      * specified {@link DomainRequirement}s.
      * <strong>NOTE:</strong> implementations are recommended to override this method if the actual secret information
      * is being stored external from Jenkins and the non-secret information can be accessed with lesser traceability
-     * requirements. The default implementation just uses {@link #getCredentialsInItem(Class, Item, Authentication, List)}
-     * to build the {@link ListBoxModel}. Handling the {@link CredentialsMatcher} may require standing up a proxy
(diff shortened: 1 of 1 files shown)


docs/consumer.adoc
```diff
@@ -281,26 +281,6 @@ We leverage the https://plugins.jenkins.io/authentication-tokens[Authentication
 This means that we can use the `AuthenticationTokens.matcher(Class)` to restrict the list of credentials to the subset that can be converted.
 Alternatively, more complex conversion contexts can be handled with `AuthenticationTokens.matcher(AuthenticationTokenContext)`
 
-* We want to let the user select the credentials to toggle the deployment state of a blue/green deployment for a completed build.
```
jglick: I do not really follow what this was about, or indeed many of the user cases listed here, which do not seem to correspond to any user functionality which was ever actually implemented.

Issue #1004, opened 2026-01-07
[card layout since 1458.v7291f840651c] prefer to have ID to be the header/title


alexsch01:
New GUI

<img width="343" height="70" alt="Image" src="https://github.com/user-attachments/assets/9d6ae42b-5989-44e8-a9f0-ae56b6d8beb5" />

I preferred it when the title was the ID (in this case ExampleID) instead of the Name / Description

---

Old GUI

<img width="1576" height="158" alt="Image" src="https://github.com/user-attachments/assets/24a350d1-511a-4589-82cb-edc8a5d2cbc0" />

Gouri-2603:
Hi! I’d like to work on this.

Before starting, I want to confirm the preferred direction:
– Would you like this implemented as a configurable option (e.g. toggle between ID vs Name/Description as the card header)?
– Or is restoring ID as the default header acceptable, with Name/Description shown secondarily?

I’m happy to prototype either approach and include screenshots for review.

alexsch01:
**[edited: Gouri-2603 is / using an AI bot]**

For me personally, restoring ID as the default header is acceptable with name/description shown secondary 

Don't know how the maintainers feel about that

Gouri-2603:
Thanks for confirming!

In that case, I’ll proceed with restoring the Credential ID as the default card header, with Name/Description shown as secondary text (similar to the previous UI behavior).

I’ll keep the change minimal and UI-only, and include before/after screenshots in the PR for easier review.

If maintainers prefer a configurable option later, we can revisit that separately.

timja:
Can we hold off for now we're still working in this area. 

There's definitely improvements to be had in the display name that goes at the top for some credentials types. Maybe the ID makes sense as the header but I don't think the linked PR as-is is enough as it removes the description. 

Yes some credentials types put the description in the display name so its duplicated information but I think better to improve that than removing the description entirely.

alexsch01:
I'll keep this open for the time being

timja:
Opened https://github.com/jenkinsci/credentials-plugin/pull/1022 sorry for the delay, the new credentials dialog changes took longer than expected to get done.

PR #1022, opened 2026-02-21
Simplify card UI

timja:
- Makes the ID as the most recognisable primary element
- Include the name as a secondary element
  - Hidden if the name is the same as the ID or the description as they are shown elsewhere
  - Bit of a hack: Remove the description from the name as its shown elsewhere

I can see why you would want to include the description in the `c:select` dropdown with the name, which is why some credentials have it in their name provider, but for the credentials list its not great so I've stripped it out.

Fixes https://github.com/jenkinsci/credentials-plugin/issues/1004

(Similar should be done on the view credentials page, I'll do that separately)

### Testing done

Tested various credentials

## Store

Before:

image

After:

image

## View credentials action

Before:

image

After:

image

### Submitter checklist

### DIFF
--- a/src/main/resources/com/cloudbees/plugins/credentials/common/card.css
+++ b/src/main/resources/com/cloudbees/plugins/credentials/common/card.css
@@ -49,6 +49,7 @@
 .credentials-card__inner--split {
     display: grid;
     grid-template-columns: 350px 1fr;
+    min-height: unset;
 }
 
 .credentials-card__details {


alexsch01:
Looks great!

janfaracik:
image

Spacing is a little broken with this

timja:
Fixed:
image

alexsch01:
Great change!! Thank you

Issue #986, opened 2025-12-21
Introduce dialogs for credentials actions


timja:
### What feature do you want to see added?

Instead of individual pages for:
* [ ] Move credentials to different domain
* [ ] Delete
* [x] New credentials
* [ ] New domain
* [ ] Update domain
* [ ] Update credentials

Use [dialogs](https://weekly.ci.jenkins.io/design-library/dialogs/) for updating these.

I think this should be fairly straightforward using the form approach.
Might be a bit more complicated on the main credentials page where all the credentials are a list as we probably don't want to include a form for every single one on the page for performance reasons.

### Upstream changes

Best done after https://github.com/jenkinsci/credentials-plugin/pull/984 lands (or branched off from it)

### Are you interested in contributing this feature?

Yes

timja:
cc @janfaracik @mawinter69

timja:
The delete one that has been created is currently taking you to the store after deletion rather than staying on the same page which isn't ideal

PR #1024, opened 2026-03-16
Convert actions into dialogs

timja:
Fixes https://github.com/jenkinsci/credentials-plugin/issues/986

This migrates all actions a user can take to a dialog rather than sub-pages.
It will simplify the new manage Jenkins layout PR: https://github.com/jenkinsci/credentials-plugin/pull/981

This was extensively written by AI. I have manually tested it myself and fixed up some minor things.
I've reviewed every line of code and cleaned it up a bit.

### Testing done

I've clicked through all the flows, they look good to me.

- [x] ATH
  - https://github.com/jenkinsci/acceptance-test-harness/pull/2643
- [x] Validate folders
- [x] Validate user credentials
- [x] Check i18n

### Submitter checklist

### DIFF
--- a/src/main/resources/lib/credentials/select/select.js
+++ b/src/main/resources/lib/credentials/select/select.js
@@ -263,9 +263,13 @@ window.credentials.submitDialog = function (form, errorMessage, onSuccess) {
     })
         .then(res => res.json())
         .then(result => {
-            window.notificationBar.show(result.data.message, window.notificationBar[result.data.notificationType]);
+            const notificationType = result.data.notificationType;
+            window.notificationBar.show(result.data.message, window.notificationBar[notificationType]);
             const dialog = document.querySelector(".jenkins-dialog");
             dialog.dispatchEvent(new Event("cancel"));
+            if (notificationType !== "SUCCESS") {
+                return;
+            }
             if (result.data.redirectUrl) {
                 // Resolve relative redirectUrl against the form action URL
                 window.location.href = new URL(result.data.redirectUrl, formAction).toString();


timja:
cc @janfaracik

janfaracik:
What did you build this with, looking pretty good!

I haven't looked through Java yet, but Jelly is looking good - made some adjustments. Would be good to clarify if we want to use `Configure` or `Update x` for updating domains/creds.

src/main/resources/lib/credentials/select/select.js
```diff
@@ -113,6 +113,13 @@ function navigateToNextPage(url, params) {
 
                 const form = newDialog.querySelector("form");
 
+                // Resolve relative form action against the fetch URL, not the current page URL,
+                // since the dialog HTML is inserted into the current page's DOM
+                const formAction = form.getAttribute("action");
```
timja: this simple fix allows simplifying all the relative path handling: https://github.com/jenkinsci/credentials-plugin/pull/1024/changes/178519c8a33b563e805c9e19bbbde7c5b11c0f58

src/main/java/com/cloudbees/plugins/credentials/CredentialsStoreAction.java
```diff
@@ -486,8 +488,17 @@ public HttpResponse doCreateDomain(StaplerRequest2 req) throws ServletException,
             Domain domain = req.bindJSON(Domain.class, data);
             String domainName = domain.getName();
             if (domainName != null && getStore().addDomain(domain)) {
+                if (jsonResponse) {
+                    return HttpResponses.okJSON(new JSONObject()
```
timja: quite average that this returns a 200, the core APIs here are not great, could create a Util here or add an API to core.

I've worked around it in https://github.com/jenkinsci/credentials-plugin/pull/1024/changes/e87b4a9d794d3061f733c116165652e48e6b6ef6

timja:
> What did you build this with, looking pretty good!

Claude did most of the work, just pointed it at the issue https://github.com/jenkinsci/credentials-plugin/issues/986.

Once it was done I tested it and pointed it at a file with all the issues.

The final cleanup I've done today was all me.
 
> I haven't looked through Java yet, but Jelly is looking good - made some adjustments. Would be good to clarify if we want to use `Configure` or `Update x` for updating domains/creds.

Update sounds good to me

janfaracik:
Tested, looks good to me! 🚀

PR #1029, opened 2026-03-22
Use Jenkins.checkGoodName for credential ID validation

Aadigarg111:
## Summary
- replace ad-hoc regex validation in `doCheckId` with `Jenkins.checkGoodName`
- keep duplicate-ID checks unchanged
- add tests for invalid IDs and a legacy stored ID compatibility case

## Testing
- Unable to run Maven tests in this environment (`mvn` is not installed).

### DIFF
--- a/src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java
+++ b/src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java
@@ -39,6 +39,7 @@ import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.ExtensionList;
 import hudson.Util;
+import hudson.model.Failure;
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/plugins/credentials/impl/BaseStandardCredentials.java
```diff
@@ -198,8 +199,10 @@ public final FormValidation doCheckId(@ContextInPath ModelObject context, @Query
             if (value.isEmpty()) {
                 return FormValidation.ok();
             }
-            if (!value.matches("[a-zA-Z0-9_.-]+")) { // anything else considered kosher?
-                return FormValidation.error("Unacceptable characters");
+            try {
+                Jenkins.checkGoodName(value);
```
timja: other than changing the allowed character set this does nothing its the same result as before.

When saving the credentials it needs to call checkGoodName not just when validating it

PR #254, opened 2021-12-09
Enable CD

jglick:
https://www.jenkins.io/doc/developer/publishing/releasing-cd/ https://github.com/jenkins-infra/repository-permissions-updater/pull/2241

PR #1037, opened 2026-04-06
Deleting Release Drafter config

jglick:
Amending #254, trying to see why #1036 [was not released](https://github.com/jenkinsci/credentials-plugin/actions/runs/24023753700/job/70057780278).

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,1 +1,0 @@
-_extends: .github


RanjitCS (approved):

jglick:
https://www.jenkins.io/doc/developer/publishing/releasing-cd/#configure-release-drafter FTR

jglick:
That seems to have worked. Perhaps there was some recent incompatible change in Release Drafter? Yet https://github.com/jenkinsci/credentials-plugin/releases/tag/1499.va_f0e811253a_6 was released more recently than the last bump to the pinned workflow, https://github.com/jenkins-infra/github-reusable-workflows/pull/59. I guess we will see if the same problem affects any other plugins which retain a vestigial RD config file. CC @timja @MarkEWaite

MarkEWaite:
> I guess we will see if the same problem affects any other plugins which retain a vestigial RD config file.

Yes.  Release drafter v7 no longer accepts the `_extends: .github` form.  It needs a more precise specification or can be removed for repositories that have CD enabled.  I've been removing the configuration file on any repository that has enabled CD because the instructions to enable CD say that the configuration is not needed.

jglick:
> Release drafter v7 no longer accepts the `_extends: .github` form.

I see. That does not explain why this only broke releases in this plugin today and not a week ago, though.

MarkEWaite:
> I see. That does not explain why this only broke releases in this plugin today and not a week ago, though.

The upgrade of the reusable workflow for CD made the switch from release drafter v6 to release drafter v7 in release [1.8.2](https://github.com/jenkins-infra/github-reusable-workflows/releases/tag/v1.8.2) four days ago.

jglick:
Ah because it is picking up the latest tag, not trunk commit.

So judging by [this query](https://github.com/search?type=code&q=org%3Ajenkinsci+NOT+is%3Aarchived+%22_extends%3A+.github%22+path%3A.github%2Frelease-drafter.yml+NOT+%22tag-template%22), a _lot_ of plugins (and other components like JTH) are now unable to be released. Maybe the v7 upgrade should be reverted until we have a handle on the situation @timja?

